PR #63, opened 2023-09-11
Add trivial fma impl

mborland:
Closes: #62


mborland:
Todo is marked on implementing the rounding modes of add, sub, mul, div. Since this is the classic example that breaks in floating point we'll know quickly if the rounding modes aren't working.

PR #67, opened 2023-09-11
Add CodeCov root-directory YAML

ckormanyos:


ckormanyos:
The purpose of this PR is to get CodeCov running with the Bot and baseline on future PRs and maybe also to add a few more files into the coverage run(s).

Cc: @mborland

mborland:
There is a file in the root right now called `.codecov.yaml`. You can overwrite that one since it clearly didn't work.

ckormanyos:
Hi Matt (@mborland) I'm still not sure if we have the automation running exaclty right.
We might have to merge this and then see.
Also you now need to add the _codecov_ job as a required job in Branch Protection I believe...

mborland:
> 

I added codecov patch as required. Project is unmarked because I don't think we are at the set threshold yet, and didn't want to block everything.

ckormanyos:
> added codecov patch as required

Cool.

I'm still a bit foggy on how all this works. I thought I reduced the tolerances to such a relaxed state in root-codecov.yml that it might be OK. Now I see that codecov/patch is _expected_ for quite a while and the Bot has not commented on this PR. I kind of forgot what to do at this point, ...

ckormanyos:
>  a bit foggy on how all this works.

But it does, in fact, work. I'll merge this for now with the goal of increasing the coverage as we plow on ahead...

Issue #65, opened 2023-09-11
Increase code coverage and maybe add quality gate


ckormanyos:
We have multiple test files each with its own `main()` subroutine. It might actually be straightforward (I think) to increase code coverage by including multiple test files in the coverage action. Maybe we could add a quality gate, but it might be too early for this.

There are some configuration settings in CodeCov that can also be handled as part of this issue. I kind of forgot if there is a _settings_ file or what, but I'll figure it out based on some of my other projects.

When adding files to CodeCov we might need to reduce testing depth since coverage instrumented files tend to run quite slowly. But so far run-time totals have not been a huge problem.

mborland:
CodeCov says we are Missing Base Commit so it is unable to show the diff on a new PR.

ckormanyos:
> CodeCov says we are Missing Base Commit so it is unable to show the diff on a new PR.

Indeed. I saw your recent PR did not get any CodeCov comment from the Bot.

This is one of those things that I've handled before. But forgot exactly how, ... I did this when setting things up in my own repos, but conveniently forgot exactly where what is needed. I'll go through and find out, ... or try to...

mborland:
It looks like it is only updating on `main`, and PR. I think adding `develop` should fix it.

ckormanyos:
> It looks like it is only updating on main, and PR. I think adding develop should fix it.

Oh yes. You do not use the same text string for the main branch. My bad. I will fix in my running PR...

ckormanyos:
See also #67

ckormanyos:
Now cycling for $90\\%$ coverage in #69.

ckormanyos:
My recent push (silly me) discarded header coverage for decimal itself. Matt (@mborland) I'll repair this ASAP...

PR #69, opened 2023-09-11
More cover files but soon need refined method

ckormanyos:


ckormanyos:
The purpose of this PR is to add more test files to CodeCov in CI.

Matt (@mborland) I'm looking for some advice on how to improve the situation. I simply listed a few files in the present CodeCov job and manually run these in the `run` part of the step.

I guess the first rudimentary improvement would be to put the working part where all the files are compiled and executed into a shell script.

On the other hand, it is not simply a given that CodeCov should mindlessly simply run all the files in the test folder. It may very well be better to take a combination of some of these compined with a certain _edge_-_cases_ file that picks up the hard-to.reach parts of coverage. Or a combination of these ideas.

Thoughts?

mborland:
Is there someway we can glob the files in the test folder? Compile everything to file_name.exe and then run every exe? I'm not proficient enough in bash to a make that happen, but that's how the cmake tests work.

ckormanyos:
> someway we can glob the files in the test folder? 

Absolutely. CMake or (my preference) GNUmake with a so-called _pattern_ _rule_. I can work this out. In fact, I think we might _like_ this one. When I do this kind of thing, I usually maintain a separate _list_ of files independent from the GNUmake or CMake code so that files can be selectively added/removed from the suite.

One silly problem-ette is that GNUmake and/or CMake can (if files are executed within) shield the CI runner from success/fail of the individual executions. But since we have fuul/deep/multi-compiler handling of the files in the required GHA runners, this is not a big deal.

I'll set up a GNUmake and we can see how it shakes out and if we like it...

ckormanyos:
Hi Matt, I still need to find the best way to exclude the test files without excluding everything. But the recent pushes here bring us up to about $90\%$ line coverage.

ckormanyos:
I'll let it cycle through and push. Then finalize the includes/excludes or find proper code annotations in another round.

ckormanyos:
See also #65

PR #73, opened 2023-09-11
Restore lost decimal-itself cover info

ckormanyos:


ckormanyos:
Struggling and fumbling, but getting close, ...

ckormanyos:
OK Matt (@mborland) we finally have clean cover info. Take a look if I _still_ missed anything as soon as it cycles on develop. If you find any errors or missing files, please let me know...

Issue #74, opened 2023-09-12
Decimal32 and Int math


mborland:
We should be able to specialize add, sub, mul, div handling when the second operand is an int rather than trying to convert it to a decimal type.

- [x] #94
- [x] #96
- [x] #98
- [x] #113
- [x] #114
- [x] #116
- [x] #119

PR #78, opened 2023-09-12
Progress on ldexp/frexp

ckormanyos:
The purpose of this PR is to complete `frexp` and `ldexp`, add tests, more coverage and handle arguments zero, inf and NaN.

### DIFF
--- a/include/boost/decimal/cmath.hpp
+++ b/include/boost/decimal/cmath.hpp
@@ -8,13 +8,18 @@
 // TODO(mborland): Allow conversion between decimal types via a promotion system
 #include <boost/decimal/decimal32.hpp>
 #include <boost/decimal/numbers.hpp>
+#include <boost/decimal/detail/cmath/fabs.hpp>
 #include <boost/decimal/detail/cmath/fmax.hpp>
 #include <boost/decimal/detail/cmath/fmin.hpp>
+#include <boost/decimal/detail/cmath/frexp.hpp>
+#include <boost/decimal/detail/cmath/ilogb.hpp>
 #include <boost/decimal/detail/cmath/isgreater.hpp>
 #include <boost/decimal/detail/cmath/isless.hpp>
 #include <boost/decimal/detail/cmath/isunordered.hpp>
-#include <boost/decimal/detail/cmath/trunc.hpp>
+#include <boost/decimal/detail/cmath/ldexp.hpp>
+#include <boost/decimal/detail/cmath/pow.hpp>
 #include <boost/decimal/detail/cmath/remainder.hpp>
+#include <boost/decimal/detail/cmath/trunc.hpp>
 
 namespace boost { namespace decimal {
 
--- a/include/boost/decimal/decimal32.hpp
+++ b/include/boost/decimal/decimal32.hpp
@@ -173,6 +173,7 @@ private:
 
     friend constexpr auto ilogb(decimal32 a) noexcept -> int;
     friend constexpr auto frexp(decimal32 v, int* expon) noexcept -> decimal32;
+    friend constexpr auto ldexp(decimal32 v, int e2) noexcept -> decimal32;
 
     template <typename T>
     BOOST_DECIMAL_CXX20_CONSTEXPR T floating_conversion_impl() const noexcept;
@@ -2025,7 +2026,7 @@ constexpr auto ilogb(decimal32 d) noexcept -> int
 {
     const auto offset = int { detail::num_digits(d.full_significand()) - 1 };
 
-    auto e10 = int { static_cast<int>(d.full_exponent()) - detail::bias + offset };
+    auto e10 = int { static_cast<int>(d.full_exponent()) + static_cast<int>(offset - detail::bias) };
 
     if (offset == 0)
     {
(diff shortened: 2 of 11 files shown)


ckormanyos:
Hi Matt (@mborland) I have a direct, maybe silly question. Can I use exponent manipulation around [here](https://github.com/cppalliance/decimal/blob/cdc1f7c5a132a40e6419cb98c913f35b157c04ff/include/boost/decimal/decimal32.hpp#L2109)?

ckormanyos:
> takes a new biased value and converts it into the right bit pattern. Try:

I'm afraid the code you suggested fails **all** of the `frexp`/`ldexp` tests. I tried a bit of debug, but was not able to fix it easily locally. Maybe I have not comprehended the bias/offset interaction yet. Well I know I haven't, but the suggested code doesn't work.

mborland:
> > takes a new biased value and converts it into the right bit pattern. Try:
> 
> I'm afraid the code you suggested fails **all** of the `frexp`/`ldexp` tests. I tried a bit of debug, but was not able to fix it easily locally. Maybe I have not comprehended the bias/offset interaction yet. Well I know I haven't, but the suggested code doesn't work.

I added print statements and this is the result:

```
Float: 1
Dec: 9.999995e-01
test_frexp_ldexp.cpp(180): test 'result_test_frexp_ldexp_is_ok' failed in function 'int main()'
Float: 0.553722
Dec: 5.537220e-01
test_frexp_ldexp.cpp(180): test 'result_test_frexp_ldexp_is_ok' failed in function 'int main()'
Float: 0.962466
Dec: 9.624655e-01
test_frexp_ldexp.cpp(180): test 'result_test_frexp_ldexp_is_ok' failed in function 'int main()'
Float: -0.579044
Dec: -5.790445e-01
test_frexp_ldexp.cpp(180): test 'result_test_frexp_ldexp_is_ok' failed in function 'int main()'
4 errors detected.
```
Editing the exponent would have less rounding steps.

ckormanyos:
> Editing the exponent would have less rounding steps.

I must be a bit dull here, but could you spell-this-out for me. The printouts look good. I can't see what I'm missing.

mborland:
> > Editing the exponent would have less rounding steps.
> 
> I must be a bit dull here, but could you spell-this-out for me. The printouts look good. I can't see what I'm missing.

I think skipping rounding steps is pushing you out of tolerance since decimal32 has higher precision than float. 1e-7 is dec32 epsilon so all of those would be a few epsilon off. I have the same issue comparing <cmath> functions to float since dec32 will have the extra digit.

ckormanyos:
>>> Editing the exponent would have less rounding steps.

>> I must be a bit dull here, but could you spell-this-out for me. The printouts look good. I can't see what I'm missing.

> I think skipping rounding steps is pushing you out of tolerance

Ah OK. Modifying the exponent is a _kind_ _of_ multiplicatoin. I need to round after that. Is there a way I could insert a rounding there? I have not looked into the rounding mechanisms yet. I'll need help on this one.

mborland:
> > > > Editing the exponent would have less rounding steps.
> 
> > > I must be a bit dull here, but could you spell-this-out for me. The printouts look good. I can't see what I'm missing.
> 
> > I think skipping rounding steps is pushing you out of tolerance
> 
> Ah OK. Modifying the exponent is a _kind_ _of_ multiplication. I need to round after that. Is there a way I could insert a rounding there? I have not looked into the rounding mechanisms yet. I'll need help on this one.

On second thought I don't think editing the exponent will work here. Is this still supposed to be num * 2^exp? Editing the exp would result in num * 10^exp which is how I implemented `sclabn`

ckormanyos:
> I don't think editing the exponent will work here. Is this still supposed to be num * 2^exp?

Uhhh yes, of course. The radix is 10. Silly me.

ckormanyos:
So there will be one _kind_ of optimization coming later, when we have optimized mul-by-int (see also #74). When that gets in, you can simply mul-by-integral-type when e2 is less than 2^32 or 2^64. But that can wait till later. I'll finish up the tests, zero, inf and NaN then.

ckormanyos:
Hi Matt (@mborland) So this one is _mostly_ complete. I just need edge cases for `ldexp`

But I'm going to get this one in to develop sooner rather than later (even if I need to post-work on it). The reason is because I finally adapted the func-prototypes to the `cmath`-philosophy in your repo.

I was a little unsure about some of the exact prototype locations and class friendship of a template class tends to be a bit tricky. So we might have to review this.

Also, the friendship would not be needed if we added just a few public constant `get`-functions for the decimal components.

ckormanyos:
One of the cool things is, that when this cycles and gets pushed, the impls of `ldexp`, `frexp`, `ilogb` and `pow` will be (will have become) generic for all decimal types in the project.

mborland:
> I was a little unsure about some of the exact prototype locations. And the syntax for enable-if class friendship of a template class tends to be a bit tricky. So we might have to review this.
> 

That never seems to work very well.

> Also, the friendship would not be needed if we added just a few public constant `get`-functions for the decimal components.

That's true. My reasoning so far to not have public getters or setters is a regular type wouldn't have them, and I want the operation to be as seamless as a builtin type. `frexp10` is kind of a cheating way I've gotten around this since it gives you the significand and exponent exactly. I should probably start writing documentation...

ckormanyos:
All true Matt. I went ahead and merged this. Even though it's not _quite_ done.

Outstanding tasks:
- I believe I can get rid of at least one class friendship even now.
- Also I still need to handle edge-cases in `ldexp`.
- I'm not sure about the exact locatoins/forms of the `enable_if`-ed prototypes. If I fumble around the architecture, I'll surely blow it and make it worse. I would ask you to review my protos on `frexp`, `ldexp`, `ilogb`, `pow`.
- Of course there are a few other template forms of `pow` that we'll need, but I'll handle these in future PR(s).

Issue #86, opened 2023-09-12
ilogb


ckormanyos:

ckormanyos:
Handling edge-cases and tests thereof still  needed.

ckormanyos:
There was some discussion on this in #93. What is now in `ilogb` is basically equivalent to the base-10 order of the number. But this might not be the same as what is needed from `ilogb`.

PR #93, opened 2023-09-15
Edge cases and no friends

ckormanyos:
The purpose of thei PR is to examine how `<cmath>`-style funcs could be implemented without `decimal` class frienship. Mostly some constant _getter_ functions would need to be publicized. I could negotiate on publicizing the ability to actually _modify_/_write_ the sign, since unary minus could do the trick for changing sign.

Actually most of the already-written `<cmath>`-funcs could be non-friends if a few _getter_ functions were public. And then these could, subsequently be written once only and genericlly for the remaining decimal class(es).

This is not mandatory, but posed here as an idea.


ckormanyos:
Hi Matt (@mborland) this style could be a bit controversial. It's definitely your choice which or even if we should try to remove these kinds of class friendships. I can go either way...

But it could/would really simplify `<cmath>`-style funcs.

Thoughts?

ckormanyos:
Further speculating, ... if the _masks_ used to represent `inf`, `NaN`, etc. wer scalable/specializable, then also funcs like `isnan`, `isinf` could become generic. Either template variables or specializations of functions come to mind for the masks.

Of course, `signbit` becomes generic with public `isneg`, as does `abs` and `fabs` uses `abs`, and so forth.

ckormanyos:
The two topics of reducing class friendships and going more generic on `<cmath>`-funcs are actually mostly independent, as the latter can also be done while maintaining the private details and using friendship. But then you need all those forward declaratikions and confusing template friendships. It's a mess, but doable.

mborland (commented):
I still don't think we should publicly expose any member function since this is supposed to replicate a built-in type. The TR says ultimately we should define `std::is_class<decimal>` to `false`. `frexp10` will give you the equivalent of `full_significand()` and `biased_exponent()` in one shot.

As for making the `inf`, `nan`, etc. masks as template variables I think that's smart. You can see I have the giant list of masks that I used trying to assemble the initial construction. Those names are all going to collide as soon as we add other types.

ckormanyos:
> don't think we should publicly expose any member function

Understood. I'll refactor and publish when ready. Keep the private parts _private_.

ckormanyos:
Hi Matt (@mborland) although this is green and private is back to private, I found some issues I'd like to clear up.

I found discrepancies between our impls of `ilogb` and `frexp10` and left some `TODO(mborland)` comments for you to check. I actually think we might have both been wrong in exponent manipulation or I was mistaken or you. But we disagreed so something must be wrong.

Second point naming. I ran into trouble with the name `exp`. Is it a base-2 exponent, a base-10 exponent, the struct member `exp` or will we even be confused about the `<cmath>` function `exp()`?

We need a naming convention such as:

- Use `expptr` or `expval` for base-2 non-members.
- Use `exp10ptr` or `exp10val` for base-2 non-members.
- For the struct, use something like `my_exp`, `my_sig`, `my_neg` or trailing underscore(s).

But mostly have a sensible rational in naming. Make sure it makes good common sense and is uniquely identifyable. And then we need to use it consistently.

mborland:
> Hi Matt (@mborland) although this is green and private is back to private, I found some issues I'd like to clear up.
> 
> I found discrepancies between our impls of `ilogb` and `frexp10` and left some `TODO(mborland)` comments for you to check. I actually think we might have both been wrong in exponent manipulation or I was mistaken or you. But we disagreed so something must be wrong.


`ilogb` should just return `num.full_exponent()` which is the unbiased exponent (I can rename that one to `unbiased_exponent()`). `frexp10` returns a base-10 significand, and a biased exponent so the user doesn't need the member functions to get those.

> 
> Second point naming. I ran into trouble with the name `exp`. Is it a base-2 exponent, a base-10 exponent, the struct member `exp` or will we even be confused about the `<cmath>` function `exp()`?
> 

The exponent and significand are in base-10 everywhere except for where bit-fiddling occurs.

> We need a naming convention such as:
> 
> * Use `expptr` or `expval` for base-2 non-members.
> * Use `exp10ptr` or `exp10val` for base-10 non-members.
> * For the struct, use something like `my_exp`, `my_sig`, `my_neg` or trailing underscore(s).
> 
> But mostly have a sensible _rationale_ in naming. Make sure it makes good common sense and is uniquely identifyable. And then we need to use it consistently.

That's fair. Since base 10 is the default for everything I would callout variables that are not (e.g. `exp2ptr`). I'll add trailing underscores to the struct members.

ckormanyos:
Maybe I'm missing something, but in comments like [these](https://github.com/cppalliance/decimal/blob/84d1643f571431c137d152b347eba7231b6e9392/test/test_decimal32_cmath.cpp#L205) I basically removed your test. Either I'm not getting something, as is so often the case, or something is up.

Shouldn't `frexp10(10^11, &n)` result in a return value of 1.0 and `n=11`?

ckormanyos:
My PR might have msessed up your `frexp10` imple in `decimal32`. And I'm not sure how to straighten this all out. But still something seems like it might have been wron in tests or impl or both?

ckormanyos:
> returns a base-10 significand, and a biased exponent

Shouldn't `frexp10` _return_ a number between 0.5 and 1.0 and fill `*expptr` with the full base-10 exponent? That would be similar to the base-2 `frexp`?

mborland:
> Maybe I'm missing something, but in comments like [these](https://github.com/cppalliance/decimal/blob/84d1643f571431c137d152b347eba7231b6e9392/test/test_decimal32_cmath.cpp#L205) I basically removed your test. Either I'm not getting something, as is so often the case, or something is up.
> 
> 
> 
> Shouldn't `frexp10(10^11, &n)` result in a return value of 1.0 and `n=11`?


I thought it would be more useful to return the signed normalized significand as an integer.

ckormanyos:
>> I'm not getting something

> I thought it would be more useful to return the signed normalized significand as an integer

Ummm, ... OK. But I might think the _name_ `frexp10` would make some folks think of the well-known `frexp` for base-2. I think you _might_ consider simply making up an unrelated name that indicates exactly what the functionality of `frexp10` does?

ckormanyos:
Sometimes I use name(s) like `get_rep` for this to expose the representation? But that does not really make sense here either.

mborland:
> >> I'm not getting something
> 
> 
> 
> > I thought it would be more useful to return the signed normalized significand as an integer
> 
> 
> 
> Ummm, ... OK. But I might think the _name_ `frexp10` would make some folks think of the well-known `frexp` for base-2. I think you _might_ consider simply making up an unrelated name that indicates exactly what the functionality of `frexp10` does?


I can add better docs. Frexp returns a normalized significand and an exponent so this is very similar. Just everything is in base 10.

ckormanyos:
> Since base 10 is the default for everything I would callout variables that are not (e.g. exp2ptr). I'll add trailing underscores to the struct members

OK. Thank you.

I do, however, still object to any local variables being called `exp`, such as the `<cmath>` functions is named. Could we please agree on something else, such as `expval` for base-10 exponent values?

ckormanyos:
OK. I've restored things according to your responses Matt (@mborland). I put the `ilogb` back into the form that I has originally written, and restored its class friendship. But I believe this implementation of `ilogb` itself is not consistent with one of your descriptions above. It is, however, used in `frexp`. So if we correct or modify `ilogb`, we will need to adjust `frexp` accordingly.

Could you please at a time of your convenience, review my impls of `ilogb` and `frexp`?

When this cycles though, I'm going to merge to develop. It's in a better state than before the PR. But we still have some defs and adjustments probably outstandeing in `ilogb` and/or `frexp`.

Then I'll move on to other things in the project basis, such as trailing return type.

ckormanyos:
I'll leave #86 open even after this PR.

Issue #103, opened 2023-09-16
Setup clang-tidy run in CI


ckormanyos:
Setup a clang-tidy run in CI. Use _most_ common rule sets. Run in GNUmake on Ubuntu native.

ckormanyos:
Cycling in #105

PR #105, opened 2023-09-16
First draft at clang-tidy CI job

ckormanyos:


ckormanyos:
So this is shaping up well. clang-tidy is running on CI.

It is too early for a quality gate on this. So I'll just print out 1 single representative tidy result.

I used a somewhat relaxed rule set. With thes, the first indication estimates something like ${\lt}2,000$ clang-tidy messages. This is actually OK and not bad for a pretty good project up front at the outset. At first you kind of grumble, but these rules make a lot of sense and adnhering to them really helps quality.

Cc: @mborland

ckormanyos:
See also #103

ckormanyos:
Matt feel free to add the "tidy" job to the required checks for branch protection, ... PR, merge to develop, etc. in GHA settings. I don't have the creds for that.

PR #106, opened 2023-09-16
Put a word count on tidy warnings

ckormanyos:


ckormanyos:
OK so the [word count](https://github.com/cppalliance/decimal/actions/runs/6208159875/job/16854054642#step:6:17) on warnings gives about ${\sim}1,500$ seemingly somewhat legit messages on a rather relaxed rule set.

Issue #102, opened 2023-09-16
Modernize use trailing return type all over


ckormanyos:
This issue tracks modernization to use trailing return type for subroutines all  over the project.

vinniefalco:
Oh god, really? What a waste of effort...

ckormanyos:
> Oh god, really? What a waste of effort...

That's what I thought at first. But having faced comments regarding non-modern code from several other open source projects, I just said, hey, why not use that modernization? It's in the language.

I know some developers resist the trailing return type, and it doesn't really add any actual value.

So this is not really my project. I did some trailing return types, but ended up deactivating (temporarily) this particular clang-tidy rule.

It's up to Matt (@mborland) if he'd like to persist with it, which I do recommend doing in order to simply be less vulnerable to criticism.

ckormanyos:
To be clear, I will not argue, neither for nor against this one, as I consider myself a guest in this project. Offline chats with Matt went in this direction and Matt can formally decide.

PR #109, opened 2023-09-17
Begin clang-tidy round one

ckormanyos:


ckormanyos:
The purpose of this PR is to handle clang-tidy messages round-one. These include a subjective subset of findings in the most important files. When doing so also handle trailing return type(s).

Relevant for #109 and #102

ckormanyos:
OK. So while this one cycles, I did stuff like:

- Remove (or handle) a bunch of super-trivial cases.
- Cast table entries and use fixed-width macros
- Remove a few more high-level rule(s) from tidy config.

When this iteration on this PR cycles though, I'll probably go ahead and merge.

The remaining tidy messages (other than trailing return type) are getting more and more into the realm of having _actual_ technical content and quality-improvement-value. So we eill need to look into these further with a bit more sophistication.

In the next rounds if it's feasible, maybe we could go for even higher code-coverage if we would like to improve confidence while treating trickier tidy messages.

Cc: @mborland

ckormanyos:
So the total gross _bulk_ of tidy message text has been reduced from an initial ${\approx}20MB$ to about ${\approx}2MB}.

The message count in the trivial `quick.cpp` test file is down from ${\sim}1,500$ messages to ${\sim}300$

That's more like it. And if this cycles green, this project in the tidy realm is now in workable range.

PR #112, opened 2023-09-18
Change handling of ilogb

mborland:
Add the non-finite paths and changes the return to be the unbiased exponent:

`
Extracts the value of the unbiased exponent from the floating-point argument num, and returns it as a signed integer value. The library provides overloads of std::ilogb for all cv-unqualified floating-point types as the type of the parameter num. (since C++23)
`

Closes: #86 

@ckormanyos


ckormanyos:
> and changes the return to be the unbiased exponent

Thank you Matt (@mborland) for clearing this up. I was quite confused about this and it will definitely help as we move on to `log10`, `log`, etc.

PR #118, opened 2023-09-18
Preliminary impl log

ckormanyos:
The purpose of this PR is to sketch out a preliminary implementation of log, scalable to `decimal32` and `decimal64` (although only 32 is available at the moment).


include/boost/decimal/decimal32.hpp
```diff
@@ -2368,7 +2377,9 @@ constexpr auto floord32(decimal32 val) noexcept -> decimal32
     }
 
     auto new_sig {val.full_significand()};
-    auto abs_exp {std::abs(val.biased_exponent())};
+
+    auto abs_exp = val.biased_exponent(); if (abs_exp < 0) { abs_exp = -abs_exp; }
+
```
mborland, line 2382: This feels like a more verbose way to the same end.

include/boost/decimal/detail/cmath/ilogb.hpp
```diff
@@ -31,9 +31,13 @@ constexpr auto ilogb(T d) noexcept -> std::enable_if_t<detail::is_decimal_floati
         return FP_ILOGBNAN;
     }
 
-    const auto offset { detail::num_digits(d.full_significand()) - 1 };
+    const auto offset =
+        static_cast<int>
+        (
+            detail::num_digits(d.full_significand()) - static_cast<int>(INT8_C(1))
```
mborland, line 37: This double cast adds new warning for redundant casting. `num_digits` returns an int and 1 is an int

include/boost/decimal/detail/cmath/log.hpp
```diff
@@ -0,0 +1,129 @@
...
+    }
+    else if(x > one)
+    {
+        constexpr auto two  = T { 2 };
+
+        // The algorithm for logarithm is based on Chapter 5, pages 35-36
+        // of Cody and Waite, Software Manual for the Elementary Functions,
+        // Prentice Hall, 1980.
+
+        auto exp2val = int { };
```
mborland, line 44: `int exp2val {}` seems like the cleaner option in the cases where you don't initialize the type with something besides default. That is the way I have been going.

include/boost/decimal/detail/cmath/log.hpp
```diff
@@ -0,0 +1,129 @@
...
+
+        // TODO(ckormanyos) There is probably something more efficient than calling frexp here.
+        auto g = (x > two) ? frexp(x, &exp2val) : x;
+
+        BOOST_DECIMAL_CXX20_CONSTEXPR auto value_sqrt2 = static_cast<T>(1.41421356237309504880L);
```
mborland, line 49: `boost/decimal/numbers.hpp` provides a C++14 constexpr variable template for sqrt2.

include/boost/decimal/detail/cmath/log.hpp
```diff
@@ -0,0 +1,129 @@
...
+                one / static_cast<std::uint8_t>(UINT8_C(12)),
+                one / static_cast<std::uint8_t>(UINT8_C(80)),
+                one / static_cast<std::uint16_t>(UINT16_C(448)),
+                one / static_cast<std::uint16_t>(UINT16_C(11264)),
+                one / static_cast<std::uint16_t>(UINT16_C(53248)),
+                one / static_cast<std::uint32_t>(UINT32_C(245760)),
+                one / static_cast<std::uint32_t>(UINT32_C(1114112)),
+                one / static_cast<std::uint32_t>(UINT32_C(4980736)),
+                one / static_cast<std::uint32_t>(UINT32_C(22020096)),
+                one / static_cast<std::uint32_t>(UINT32_C(96468992))
```
mborland, line 77: Redundant casts

include/boost/decimal/detail/cmath/log.hpp
```diff
@@ -0,0 +1,129 @@
...
+        {
+          result += term;
+
+          zn *= z2;
+
+          term = zn * coefs[index];
+        }
+
+        {
+            BOOST_DECIMAL_CXX20_CONSTEXPR auto value_ln2 = static_cast<T>(0.693147180559945309417L);
```
mborland, line 105: Numbers has an ln2 template

include/boost/decimal/detail/cmath/log10.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+template<typename T>
+constexpr auto log10(T x) noexcept -> std::enable_if_t<detail::is_decimal_floating_point_v<T>, T>
+{
+    // TODO(ckormanyos) Handle non-nermal varguments.
+
+    // TODO(ckormanyos) Put in a bisic check for pure powers of 10, resulting
+    // in an exact result.
+
+    BOOST_DECIMAL_CXX20_CONSTEXPR auto value_lg10 = static_cast<T>(2.30258509299404568402L);
```
mborland, line 25: Numbers has ln10.

include/boost/decimal/decimal32.hpp
```diff
@@ -1197,7 +1202,11 @@ constexpr auto mixed_equality_impl(decimal32 lhs, Integer rhs) noexcept -> bool
         }
     }
 
-    Unsigned_Integer rhs_significand {rhs < 0 ? detail::apply_sign(rhs) : rhs};
+    const auto rhs_significand =
+        static_cast<Unsigned_Integer>
+        (
+            (rhs < 0) ? static_cast<Unsigned_Integer>(detail::apply_sign(rhs)) : static_cast<Unsigned_Integer>(rhs)
+        );
```
mborland, line 1209: I added `detail::make_positive_unsigned` to avoid this situation now. All you need is:

`Unsigned_Integer rhs_significand {detail::make_positive_unsigned(res)};`

include/boost/decimal/decimal32.hpp
ckormanyos: Indeed. But I had a compiler that failed to have the `constexpr` version of `<algorithm>`'s function `std::abs()`. It was MSVC 142. So I kind of _had_ _to_ do this. SIgh.

include/boost/decimal/detail/cmath/log.hpp
ckormanyos: I was wondering which way was best/faster/better. I'll try to be consistent to your suggestion. If I miss a few, we can re-edit them.

include/boost/decimal/detail/cmath/log10.hpp
ckormanyos: Also of course. Got it. In this preliminary draft, `log10` of _far_ from being complete. There will/should-be an entirely different implementation that cares much more for pure powers of 10, etc.

include/boost/decimal/decimal32.hpp
ckormanyos: OK.

ckormanyos:
Thanks Matt. Give me a chance to handle the review comments.

include/boost/decimal/detail/cmath/log.hpp
ckormanyos: Oh the casting tussle... OK you got it.

include/boost/decimal/decimal32.hpp
mborland: I dream of a day where MSVC works like everyone else.

ckormanyos:
Hi Matt (@mborland) I believe I've now got the implementation and also handled the review comments. A _side_-_effect_ of this PR is also reducing the tidy-bulk by about 10 percent, resulting in fewer than 300 messages now.

I believe `log` is good-to-go...

mborland (approved):
Looks good to me. The only thing I can think of is throw in non-finite value tests so that we get the coverage of log(isnan) and log(isinf)

ckormanyos:
> throw in non-finite value tests so that we get the coverage of log(isnan) and log(isinf)

This is the kind of setup I really like. Not only was I missing those coverage points, but also had weaknesses between 0 and 1 and 1.5 to 2.0. This review comment got me to 100 percent line coverage, caused me to bump up one test-tolerance and actually uncovered a typo I had made in reflection for argument less than 1.

What a great CI setup!!!

Thanks Matt (@mborland)

Issue #101, opened 2023-09-15
Make fpclassify, etc generic instead of friends


mborland:
- [x] fpclassify
- [x] isfinite 
- [x] abs
- [x] ceil
- [x] floor
- [x] trunk

PR #122, opened 2023-09-19
Make additional cmath functions generic instead of friends

mborland:
Closes: #101


ckormanyos (commented):
I struggled with this too last week. Maybe I'm wrong or it's an MSVC problem but shouldn't these calls to `isfinite` be properly found via ADL?

mborland:
> I struggled with this too last week. Maybe I'm wrong or it's an MSVC problem but shouldn't these calls to `isfinite` be properly found via ADL?

It's definitely an MSVC issue since it finds the correct one, and then for some reason thinks it can use the <cmath> one too...

PR #124, opened 2023-09-19
Cover edges and hard-to-reach lines

ckormanyos:
The purpose of this PR is to hit more hard-to-reach lines and try to cover more edge cases.


ckormanyos:
This got another 3 percent of coverage. I left a few TODO(@mborland) comments that seem like need addressing.
Also I had trouble casting from `decimal32` to `uint8_t` and to `uint16_t`, we might need a few more operators there.

This is a rich topic and I'll continue driving up coverage and finding tidbits in future PRs.

PR #128, opened 2023-09-20
Function exp

ckormanyos:
The purpose of this PR is to implement and test `exp()`.


ckormanyos:
Preliminary implementation of `exp()` is in. I appreciate the method. It's not super-speed optimized but portaboe to 32/64. I also in a future round will want to squeeze the tolreances, as I'm needing something like $96{\times}{\epsilon}$ to pass. I would have hoped for better tolerance.

ckormanyos:
I will need to correct one final edge case for argument negative infinity, then `exp()` will be good to go.

PR #137, opened 2023-09-22
Improve coverage hits using Matt rng style

ckormanyos:


ckormanyos:
OK Well I fiddled around with some tests and used your RNG solution uniformly. We now reach $90\\%$ coverage. Yeah

PR #206, opened 2023-10-10
Awarenes for __cplusplus and _MSC_VER

ckormanyos:
The purpose of this PR is to figure out better ways to detect modern `__cplusplus` version(s).

Hi Matt (@mborland), my VC142 tests positive for the first PP-query in

```
#if __cplusplus <= 202002L || _MSVC_LANG <= 202002L
```

This is because MSVC empirically seems to have poor support for `__cplusplus`.


mborland (approved):
This should be fine. I am used to this working with > that I forgot than < would always be wrong with MSVC.

PR #207, opened 2023-10-10
Preliminary implementation and test acosh

ckormanyos:
The purpose of this PR is to implement and test `acosh`.


ckormanyos:
Hi Matt (@mborland) I found during implementation of `acosh` that the implementation of `sqrt` was in a preliminary form. I reworked `sqrt` in this PR, but it is, at the moment, in dire need of optimization. So let's make that as a note.

ckormanyos:
Notes: Optimize and extend tests of `sqrt`.

PR #222, opened 2023-10-11
Add comments for future improvements

ckormanyos:


ckormanyos:
Hi Matt (@mborland). It appears as though I've been forgetting to add my test files to the Jamfile.

Ooops.

Local tests have all been OK. But I might need to fiddle around with this if any weird warnings or compiler-specific behavior pops up. We need to also see if I added undue amount(s) of run-time for the entire test suite.

mborland:
> > added undue amount(s) of run-time
> 
> 
> 
> Hi Matt (@mborland) mostly this is straightforward. But there is one question. Do we already have a compiler switch to detect at compile-time UBSAN runs in CI?
> 
> 

Not at this time. We could copy the config from math trivially.

mborland:
> > Not at this time. We could copy the config from math trivially.
> 
> 
> 
> I found some kind of a PP-definition for `UBSAN` in the YAML. I'm not sure if it is _hooked_ _up_. But i'm close to getting elementary funcs working even though i flubbed up on the Jamfile for a few weeks. Just a few problems with clang 4.5, ... Close...

The minimum compilers in the CI are arbitrary. I took the one from charconv and started dropping anything that didn't immediately work. I wouldn't bend over backwards for clang 4.x. Setting 5 or 6 to minimum would be roughly equivalent to GCC 6 anyway.

ckormanyos:
> minimum compilers in the CI are arbitrary

Actually, I really like the choice of compilers.

It's probably going to go green. We have a few ugly PP-selections that I made to go green. I was a bit peeved that I forgot to add files to BJAM. Somehow, it's kind of hard to believe, but utterly Chris-like, to forgot to add all the carefully crafted new files to BJAM. But, hey, that's OK for now.

When this gets green, I'll address the ugly PP-selections that I made. We probably need a few `DECIMAL` selections for `CONSTEXPR` as opposed to using the Boost configoptions. But this is something we can tune later.

ckormanyos:
Hi Matt (@mborland) I'm down to one problem. UBSAN and/or ASAN runs time out on drone. I think I will need some help from you on this issue to either find out what's taking so long and reduce testing scope or break the tests up onto individual runners.

When I run ASAN, UBSAN or `_DEBUG`, I use a sequence of PP-decisions to reduce scope. It looks something like [this](https://github.com/cppalliance/decimal/blob/cd08973082dc73dd020478e9d84a3581a42f1f0d/test/test_frexp_ldexp.cpp#L6).

I'm not tsure if `UBSAN` is actually `#define`-d for _all_ of the UBSAN runs on both GHA as well  as drone. So we need to check that.

Anyway, I need a bit of help to get drone no longer timing out...

mborland:
> Hi Matt (@mborland) I'm down to one problem. UBSAN and/or ASAN runs time out on drone. I think I will need some help from you on this issue to either find out what's taking so long and reduce testing scope or break the tests up onto individual runners.
> 
> When I run ASAN, UBSAN or `_DEBUG`, I use a sequence of PP-decisions to reduce scope. It looks something like [this](https://github.com/cppalliance/decimal/blob/cd08973082dc73dd020478e9d84a3581a42f1f0d/test/test_frexp_ldexp.cpp#L6).
> 
> I'm not tsure if `UBSAN` is actually `#define`-d for _all_ of the UBSAN runs on both GHA as well as drone. So we need to check that.
> 
> Anyway, I need a bit of help to get drone no longer timing out...

It's interesting that the drone config is timing out at an hour, but the same config on GHA finishes in 40m. Anyway's I'll take a look.

include/boost/decimal/detail/cmath/exp.hpp
```diff
@@ -91,10 +91,15 @@ constexpr auto exp(T x) noexcept -> std::enable_if_t<detail::is_decimal_floating
             // TODO(ckormanyos) Is a Pade approximation more precise or faster?
             // And, ... how can you make it scalable if you go ahead and "Pade"-it?
 
-            constexpr coefficient_array_type
-                coefficient_table
+            #if (defined(__clang__) && (__clang__ < 6))
+            #  pragma GCC diagnostic push
+            #  pragma GCC diagnostic ignored "-Wmissing-braces"
```
mborland, line 96: Should this be `# pragma clang diagnostic ...`?

include/boost/decimal/detail/cmath/tanh.hpp
```diff
@@ -61,25 +61,39 @@ constexpr auto tanh(T x) noexcept -> std::enable_if_t<detail::is_decimal_floatin
...
+                #  pragma GCC diagnostic push
+                #  pragma GCC diagnostic ignored "-Wmissing-braces"
+                #endif
+
+                #if (defined(__clang__) && (__clang__ < 6))
+                const
+                #else
+                constexpr
+                #endif
+                auto coefficient_table =
```
mborland, line 74: Would pre-computing the values fix clang < 6 issue?

mborland:
I moved all of the configuration into `<boost/decimal/detail/config.hpp>` so it will always be brought in. It now covers UBSAN, ASAN, and TSAN. Looks like UBSAN was covered, but the ASAN would timeout.

mborland:
Drone build shows it now takes 31 mins which is about our median build time of 34 mins.

ckormanyos:
> Drone build shows it now takes 31 mins

Perfect distribution many thanks

include/boost/decimal/detail/cmath/tanh.hpp
ckormanyos: Yes. There is an expressional syntax that works for all the CI compielrs. I've got like 6 or seven tables and i modified tham all except for two. I'll eliminate these compiler switches and go `constexpr` on all of them prior to merging this.

ckormanyos:
Hi Matt (@mborland) actually now this looks good, but however, UBSAN on GHA is running about an hpour. Could we plesae split GCC12 UBSAN 32/64 and maybe have a chance to cut that down? Or is that already done here and I'm missing something?

mborland:
> Hi Matt (@mborland) actually now this looks good, but however, UBSAN on GHA is running about an hour. Could we plesae as on drone split GCC12 UBSAN 32/64 and maybe have a chance to cut that down? Or is that already done here and I'm missing something?

Should be fixed in https://github.com/cppalliance/decimal/pull/233. Both runs are now under 30 mins.

Issue #236, opened 2023-10-13
Decimal64 <cmath> friends


mborland:
- [ ] copysign
- [ ] fma
- [ ] frexp10
- [ ] scalbn
- [ ] scalbln

Issue #235, opened 2023-10-13
IEEE 754-2019 Extensions to <cmath> for Decimal64


mborland:
- [x] samequantumd64
- [x] quantexpd64
- [x] quantized64

PR #237, opened 2023-10-13
Add <cmath> functions and testing to Decimal64

mborland:
Closes: #235 
Closes: #236


mborland:
@ckormanyos this adds testing of almost all the cmath functions for decimal64. The 32-bit platforms are slow as expected but it looks like we're in a good spot. Also brings coverage back above 99%.

Issue #242, opened 2023-10-14
Cast decimal64 to built-in double returns 0 for edge bit patterns


ckormanyos:
Hi Matt (@mborland)

While adapting the `exp()` tests in prototype work using the brand new implementation of `decimal64`, I've ran across what appears to be an edge case.

For some bit patterns stored in `decimal64`, the conversion to built-in 64-bit `double` results in zero for the `double` result. This is happening in the range of $9.98{\times}10^{15}$, so right around that 53-54 bit area.

I have isolated an exact test case, as shown in its entirety below. I use crazy direct-memory-access to initialize what appears to be a valid `decimal64` with the bit pattern `0x6A3B71802C99CB39`. This results in a seemingly valid `decimal64`.

But when casting it to 64-bit built-in double, the result is zero. I included in the comments of the test program a few other bit patterns also featuring this issue.

I ran across these issues when hammering `exp()` tests and every now and then, the values of `exp(x)` for $x{\sim}36$ revealed this problem when computing the closeness to the expected answer. Further investigations showed, in fact, that the `static_cast` to `double` was getting messed up.

I originally detected on MSVC 142 and subsequently also find this on GCC-11 Ubuntu Jammy on WSL.

```cpp
#include <iostream>

#include <boost/decimal.hpp>

// cd /mnt/c/Users/User/Documents/Ks/PC_Software/Test
// g++ -w -O2 -I/mnt/c/MyGitRepos/cppalliance/decimal/include -I/mnt/c/boost/boost_1_83_0 test.cpp -o test.exe

auto main() -> int
{
  // Consider the following integral patterns potentially stored
  // in a given decimal64.

  // 0x6A3A25E507BB83D9
  // 0x6A3800D0288A63E9
  // 0x6A380BF1150D7F35
  // 0x6A3B71802C99CB39

  // Let's take the fourth one of these patterns.
  const std::uint64_t ui = UINT64_C(0x6A3B71802C99CB39);

  // And we convert it via direct-memory access to a decimal64.

  // In this particular example we assume that this number could
  // be the result of some calculation, initialization or any other
  // valid operations.

  const boost::decimal::decimal64 dec = *reinterpret_cast<const boost::decimal::decimal64*>(&ui);

  // Let's print this decimal64 value.
  std::cout << "dec: " << dec << std::endl;

  // Now let's convert this decimal64 value to built-in double.
  // ISSUE: This cast seems to result in 0.
  const double dbl = static_cast<double>(dec);

[...]

PR #243, opened 2023-10-14
Update and test exp 64-bit and Pade approx

ckormanyos:
The purpose of this PR is to extend the testing of the `exp()` function (known from `<cmath>`) to 64-bit decimal.

Also in this PR it was decided to see how the Pade approximation plays out for `exp()`, as this seems to give better precision for lower run-time cost.


ckormanyos:
The range of `exp()` tests is reduced from maximum argument $80.0$ down to $32.0$ in this PR because of the issue reported in #242.

ckormanyos:
Note to self: Look at `ldexp()` for larger argument.

ckormanyos:
This is fine enough for me. But the argument range of `exp()` tests on `decimal64` has been reduced pending resolution of the issue mentioned in #242.

PR #244, opened 2023-10-14
Extent log() tests to 64-bit

ckormanyos:


ckormanyos:
Hi Matt (@mborland) as we chatted offline, I'm in the process of identifying the biggest test blocks and reducing their depths when using sanitize/debug. I missed a few on the previous run, so let's see if this recent push makes the time cut or still times out, ...

Issue #205, opened 2023-10-10
Optimizations powers 10^n


ckormanyos:
The function `log10` has basically a naive implementation ot the moment. Right now, `log10` simply uses the existing implementation of `log`. Improvements would have some support for exact base-10 logarithms in the base-10 logarithm function.

ckormanyos:
- Improve `log10` for known powers of ten
- Improve `pow(dec, n)`, where ${n}%%10 = 0$

ckormanyos:
> Improve `log10` for known powers of ten

This is cycling in #544

PR #252, opened 2023-10-17
Pow

ckormanyos:
The purpose of this PR is to complete `pow()` and its tests.

### DIFF
--- a/include/boost/decimal/detail/cmath/exp.hpp
+++ b/include/boost/decimal/detail/cmath/exp.hpp
@@ -10,6 +10,7 @@
 #include <type_traits>
 
 #include <boost/decimal/fwd.hpp> // NOLINT(llvm-include-order)
+#include <boost/decimal/detail/cmath/impl/pow_impl.hpp>
 #include <boost/decimal/detail/type_traits.hpp>
 #include <boost/decimal/numbers.hpp>
 
@@ -87,7 +88,7 @@ constexpr auto exp(T x) noexcept -> std::enable_if_t<detail::is_decimal_floating
                 }
                 else
                 {
-                    result *= pow(T { 2, 0 }, nf2);
+                    result *= detail::pow2<T>(nf2);
                 }
             }
         }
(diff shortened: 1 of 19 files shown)


ckormanyos:
This one still needs a tad of work since `pow()` has all kinds of special cases and values with its two input params. I'll need another day or two on the mighty `pow()`...

Cc: @mborland

ckormanyos:
> Are you going to keep the other implementation which allowed for Integer powers?

Yes in fact I believe there is an integer power-specified function in the standard. It's like 23 or then again `constexpr` in 26. So there are a couple overloads. We might need review before all is said and done...

grafik

ckormanyos:
But now I'm not sure if the integral power is _until_ C++11 or after, or what happens after 11 to the integral power?

Cc: @mborland

ckormanyos:
>  not sure if the integral power

It the integral power is no longer standard, then we could _make_ _up_ a function. I do this all the time but it's kind of uncomfortable to leave the confines of the standard. Anyway, if I make one up, then I usually call it `pown()` and use the ladder method to compute the powers...

mborland:
> But now I'm not sure if the integral power is _until_ C++11 or after, or what happens after 11 to the integral power?
> 
> 
> 
> Cc: @mborland 

It got replaced by that template overload which allows any arithmetic type to the power of any arithmetic type. In our case I think a template for decimal to the power of a decimal and a template for decimal to an integral power would be it.

ckormanyos:
> It got replaced by that template overload which allows any arithmetic type to the power of any arithmetic type.

Coding delight, ...

> In our case I think a template for decimal to the power of a decimal and a template for decimal to an integral power would be it.

OK, cool. Fits perfectly for me. That's what I had in my original scope of implementation. Let me get the implementation, tests and special parameters right, then we can see if it's _enough_.

ckormanyos:
OK so I finally got `pow(dec, Integral)` and did some refactoring. Now I just need to handle the special cases of `pow(dec, dec)`, then that's all...

mborland:
An optimization step we could take down the is directly manipulating the exponent for powers of ten.

ckormanyos:
> optimization step [...] is directly manipulating the exponent for powers of ten

Nice. I'll note that. See also (renamed) #205.

Issue #263, opened 2023-10-19
Bitwise logical operators


mborland:
Like AVX we can guarantee we know what the internal representation

PR #267, opened 2023-10-19
Add bitwise logical operators

mborland:
Closes: #263


mborland:
@ckormanyos I think this confirms that next week I need to redo the basis for the types and constructors. GCC won't let you memcpy which wouldn't be an issue if the basis of the type was just an integer instead of a bit field. Hopefully it'll bring some performance improvements because normally indexing a bit field isn't the fastest operation.

Issue #141, opened 2023-09-23
tgamma


ckormanyos:

ckormanyos:
Implementation plan:
  - Scale argument down.
  - Use series approximation and rescale result.
  - Handle non-normal arguments.

PR #284, opened 2023-10-24
tgamma and its tests

ckormanyos:


include/boost/decimal/detail/cmath/cos.hpp
```diff
@@ -51,7 +51,7 @@ constexpr auto cos(T x) noexcept -> T
         // | 2 | -sin(r) | -cos(r) |  sin(r)/cos(r) |
         // | 3 | -cos(r) |  sin(r) | -cos(r)/sin(r) |
 
-        constexpr auto my_pi_half = numbers::pi_v<T> / 2;
+        BOOST_DECIMAL_CXX20_CONSTEXPR auto my_pi_half = numbers::pi_v<T> / 2;
```
mborland, line 54: I don't think this should get bumped up to C++20 just because MSVC 14.1 is non-compliant. What about:

```
#if defined(_MSC_VER) && _MSC_VER == 1910
const auto my_pi_half = numbers::pi_v<T> / 2;
#else 
constexpr auto my_pi_half = numbers::pi_v<T> / 2;
#endif
```

ckormanyos: Doesn't BOOST_ something tell MSVC? I could use raw `_MSC_VER` but I thought config had some of that stuff? Perhaps I'm mistaken...

ckormanyos: Also, if I do this, I tend to use `_MSC_VER < 1920` which does an excellent job picking up all of 141...

mborland: I am pretty sure that `BOOST_MSVC` is just an alias to `_MSC_VER`. Either way if we use raw `_MSC_VER` we can maintain standalone support.

ckormanyos: It's cycling. If it works, we can tune it. Let's see if it goes green...

ckormanyos:
Fix #293 and fix #141

PR #286, opened 2023-10-26
Remove cxxstd=2b for clang < 16

mborland:
Avoids: https://stackoverflow.com/questions/77164931/clang-error-on-standard-library-depending-on-std-c2b-version-and-warning-wh

@ckormanyos you may be hitting this in your other libraries. Fixed it this morning on charconv too.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -172,31 +172,31 @@ jobs:
               - clang-11
           - toolset: clang
             compiler: clang++-12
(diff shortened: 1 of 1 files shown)


ckormanyos:
> @ckormanyos you may be hitting this in your other libraries

Yes. Thank you for pointing out this phenomenon, Matt (@mborland). Several things seem to be going haywire on some GHA runners and I'm refactoring those other projects too.

By the weekend, I should have time to get back to some contributions on decimal.

PR #315, opened 2023-11-13
Improve standalone support

mborland:
Improves support for bare metal platforms


ckormanyos:
What a nightmare, MSVC does not _know_ how to count leading zeros. I have a software impl somewhere, but it's all over the place too.

PR #316, opened 2023-11-13
Bare metal

ckormanyos:
Hi Matt (@mborland) as reported in another chat, I have a very small part of decimal running now on a bare-metal ARM(R) Cortex(R) M4 in hard real time.

- I can report on some highlights as this PR progresses.
- I'll also put a tiny QEMU ARM(R) Cortex(R) M4 run in CI.

Basically plug-n-plan, but you need a configuration option `BOOST_DECIMAL_DISABLE_IOSTEAM`, which you are welcome to change at will.

The embedded compiler detects a bunch of conversion  and sign-conversion warnings, which I can handle one after the other.


ckormanyos:
As we discussed previously, `core/bits.hpp` pulls in `BOOST_STATIC_ASSERT` which breaks the very nice standalone feature. Workaround is highly recommended to regain standalone.

mborland:
> As we discussed previously, `core/bits.hpp` pulls in `BOOST_STATIC_ASSERT` which breaks the very nice standalone feature. Workaround is highly recommended to regain standalone.

Should be fixed by: https://github.com/cppalliance/decimal/pull/315

ckormanyos:
I linked with nano-specs and can give a size indicatoin of something around 20k ROM consumption. The project pulls in the floating-point library, so that is not entirely _on_ the Decimal implementation.

A bunch of standard C-LIB stuff is pulled in, but that's easy enough to stub, but irrelevant for this example, as it will not run anyway.

Here is an extended MAP file.
[ref_app_cppfilt.txt](https://github.com/cppalliance/decimal/files/13333881/ref_app_cppfilt.txt)

A picture is below:

grafik

mborland:
> I linked with nano-specs and can give a size indicatoin of something around 20k ROM consumption. The project pulls in the floating-point library, so that is not entirely _on_ the Decimal implementation.
> 
> A bunch of standard C-LIB stuff is pulled in, but that's easy enough to stub, but irrelevant for this example, as it will not run anyway.
> 
> Here is an extended MAP file. [ref_app_cppfilt.txt](https://github.com/cppalliance/decimal/files/13333881/ref_app_cppfilt.txt)
> 
> A picture is below:
> 
> grafik

There's a ton of static data here for float conversions: https://github.com/cppalliance/decimal/blob/develop/include/boost/decimal/detail/fast_float/significand_tables.hpp

ckormanyos:
> a ton of static data here for float conversions

I'm actually pleasantly happy with the whole _metal_ _experience_ and recommend no change whatsoever in those areas.

ckormanyos:
So we expected this, but `Decimal` is software-intensive as an emulated, base-10 type on binary machines. And it is expectedly slow.

The target is clocked at $168 MHz$ and has 5 wait states on flash acess with small amounts of pipelining and I/D-caching. I suspect a throughput of more than $100$ operations per ${\mu}s$.

The scope shows the `decimal64` timing for the `exp()` calculation is about $8ms$, resulting in probably something like

$$
8000 {\mu}s {\times} 100 ops/{\mu}s {\approx} 8^5 ops
$$

decmial64_exp

ckormanyos:
Hi Matt (@mborland) here is the first benchmark we discussed on ARM(R) Cortex(R) M4.

Using 3 template types:

- built-in `double` (with no FPU)
- software-emulated `soft_double`
- `decimal64`

## Setup Description

Use a common 32-bit ARM(R) bare-metal microcontroller
STM32F446 ARM(R) Cortex(R) M4 at $168~MHz$
GCC11, arm-none-eabi, float-abi=soft, -O2

Calculation exp() via Pade approximation approx. 15 decimal digits.
Code-size includes about 2k for startup, clock-init, skinny MCAL and cooperative time-scheduler.


## Target-Specific Compiler Flags

```
-O2 -mcpu=cortex-m4 -mtune=cortex-m4 -mthumb -mfloat-abi=soft -finline-functions -finline-limit=128 -mno-unaligned-access -mno-long-calls
```

## Benchmark Results

| Type                           |  runtime [us] |  relative  |    code-size [kb] |
|--------------------------------|---------------|------------|-------------------|                               
| double (built-in, no FPU)      |  22           |   1.0      |       5.6         |
| ::math::softfloat::float64_t   |  27           |   1.2      |       8.5         |
| boost::decimal::decimal64      |  2800         |   130      |       34          |

ckormanyos:
The source code for the above-mentioned benchmark can be found
[here](https://github.com/ckormanyos/real-time-cpp/blob/bare_metal_decimal/ref_app/src/app/benchmark/app_benchmark_non_std_decimal.cpp),
in the `bare_metal_decimal`-branch of [real-time-cpp](https://github.com/ckormanyos/real-time-cpp).

ckormanyos:
Matt, I just requested you for review on this.

Personally, I'm totally happy with the bare-metal adaption. I was able to cut out _ALL_ unecessary C-LIB stuff with the right compiler switches and removal of `<cassert>`.

The latter will be controversial, but perhaps we could enable/disable asserts via configuration option. Anyway this runs perfectly well _on_ _the_ _metal_. The performance will be an issue for down the road. I gave up on 8-bit compilation since the compiled RAM/ROM significantly exceeded the ARDUINO capacity. I'll look into that later.

Anyway, with the right config options this one is either ready or soon to be ready to go, ...

mborland (commented):
Do your standard libraries include `<iostream>`, `<cwchar>`, etc.? If not I think something like this could get added to the top of config to automatically disable IO streams: https://github.com/boostorg/math/blob/develop/include/boost/math/tools/is_standalone.hpp

ckormanyos:
> Do your standard libraries include `<iostream>`, `<cwchar>`, etc.?

This is a good questoin. And, honestly, some of my offline chatter has been a bit confusing. This PR is intended to handle bare-metal platforms which _do_, in fact, have a complete C++ standard library. These will include `<iostream>`, `<ios>`, `<cassert>` and the like.

The idea is to contrive a (probably user-set) configuration option which will disable the _pulling_-_in_ of all such heavyweights.

I'm not sure how to best do this. But asking is a compiler `__has`-I/O-streaming will not work because the ARM/RISC, etc. compilers will have them.

mborland:
A user defined macro is fine then. I do think we should keep the asserts and make those into a macro. I’ve hit a few of them along the way in development so I’d prefer to keep them.

ckormanyos:
> keep the asserts and make those into a macro

Sounds fine.

Please relay to me the name of the macro. I can re-activate the asserts under the macro and also associate the empty-macro with another macro that disables I/O streaming and asserts all in one.

How about:

- `BOOST_DECIMAL_DISABLE_IOSTREAM` to disable I/O streaming
- `BOOST_DECIMAL_ASSERT(expr)` for the assert macro

What I can't figure out is how to make a combi-macro that wraps up together no-assert/no-I/O-streaming for the metal... Ideas?
-

mborland:
> > keep the asserts and make those into a macro
> 
> Sounds fine.
> 
> Please relay to me the name of the macro. I can re-activate the asserts under the macro and also associate the empty-macro with another macro that disables I/O streaming and asserts all in one.
> 
> How about:
> 
> * `BOOST_DECIMAL_DISABLE_IOSTREAM` to disable I/O streaming
> * `BOOST_DECIMAL_ASSERT(expr)` for the assert macro
> 
> What I can't figure out is how to make a combi-macro that wraps up together no-assert/no-I/O-streaming for the metal... Ideas?

Something like:
```
#ifndef BOOST_DECIMAL_DISABLE_CASSERT
#  define BOOST_DECIMAL_ASSERT(x) assert(x)
#endif

#ifdef BOOST_DECIMAL_DISABLE_CLIB
#  ifndef BOOST_DECIMAL_DISABLE_IOSTREAM
#    define BOOST_DECIMAL_DISABLE_IOSTREAM
#  endif
#  ifndef BOOST_DECIMAL_DISABLE_CASSERT
#    undef BOOST_DECIMAL_ASSERT
#    define BOOST_DECIMAL_ASSERT(x) 
#  endif
#endif

```

ckormanyos:
>> What I can't figure out is how to make a combi-macro

> Something like:

OK. Got it. I'll implement and test something like you show above.

As a final step, I will put together a little bare-metal build+QEMU-run and put a build-detection gate at under 20k ROM or something. This is what I usually do to ensure that no heavyweight stuff ultimately creeps in, ...

Need another couple days to wrap all this up...

ckormanyos:
Hi Matt (@mborland) this one should be good to go.

- Added `BOOST_DECIMAL_DISABLE_IOSTREAM`
- Added `BOOST_DECIMAL_DISABLE_ASSERT`
- Added `BOOST_DECIMAL_DISABLE_CLIB`
- Added a bare-metal QEMU run in CI on ARM(R) Cortex(R)-M4
- Restored all the original asserts using new macros.

There is a new mandatory run in CI named `benchmark_single-stm32f429-qemu`.

ckormanyos:
Hi Matt (@mborland) actually I need to finish disabling some C-LIB stuff. This one is not quite ready yet.

ckormanyos:
Hi Matt please take a look at this. ome of the config options are used in a rather widespread fashion. I could strengthen the QEMU success chek in Ci and i'll do that in a future activity.

But for the moment, from my side, this bare-metal thing is good to go. The code isze is super clean now that the C-LIB stuff is fully configures out - around 11kb code on the metal.

And, as mentioned above, there is anew mandatory CI job for the CM4 QEMU run in `metal.yml`.

include/boost/decimal/decimal128.hpp
```diff
@@ -47,6 +35,21 @@
...
+#include <cmath>
+#include <cstdint>
+#include <cstring>
+#if !defined(BOOST_DECIMAL_DISABLE_IOSTREAM)
+#include <cwchar>
+#include <iostream>
+#endif
+#include <limits>
+#if !defined(BOOST_DECIMAL_DISABLE_IOSTREAM)
+#include <sstream>
```
mborland, line 49: These 2 can be combined, and I think should be moved to the end of the include list since that will be less ugly.

include/boost/decimal/decimal32.hpp
```diff
@@ -46,6 +34,18 @@
...
 
+#include <cerrno>
+#include <climits>
+#include <cmath>
+#include <cstdint>
+#include <cstring>
+#if !defined(BOOST_DECIMAL_DISABLE_IOSTREAM)
+#include <cwchar>
+#include <iostream>
+#endif
```
mborland, line 45: Move to end

include/boost/decimal/detail/apply_sign.hpp
```diff
@@ -34,7 +34,7 @@ template <typename Integer, typename Unsigned_Integer = detail::make_unsigned_t<
           std::enable_if_t<detail::is_signed_v<Integer>, bool> = true>
 constexpr auto make_positive_unsigned(Integer val) noexcept -> Unsigned_Integer
 {
-    return static_cast<Unsigned_Integer>(val < static_cast<Integer>(0) ? apply_sign(val) : val);
+    return static_cast<Unsigned_Integer>(val < static_cast<Integer>(static_cast<std::int8_t>(INT8_C(0))) ? apply_sign(val) : val);
```
mborland, line 37: Is this necessary? We are already casting 0 to the target type.

include/boost/decimal/detail/bit_layouts.hpp
```diff
@@ -91,7 +91,7 @@ struct IEEEl2bits
 
 struct IEEEl2bits
 {
-#if BOOST_CHARCONV_ENDIAN_LITTLE_BYTE
+#if (defined(BOOST_CHARCONV_ENDIAN_LITTLE_BYTE) && (BOOST_CHARCONV_ENDIAN_LITTLE_BYTE != 0))
```
mborland, line 94:
```suggestion
#if (defined(BOOST_DECIMAL_ENDIAN_LITTLE_BYTE) && (BOOST_DECIMAL_ENDIAN_LITTLE_BYTE != 0))
```

include/boost/decimal/detail/bit_layouts.hpp
```diff
@@ -112,7 +112,7 @@ struct IEEEl2bits
 
 struct IEEEbinary128
 {
-#if BOOST_CHARCONV_ENDIAN_LITTLE_BYTE
+#if (defined(BOOST_CHARCONV_ENDIAN_LITTLE_BYTE) && (BOOST_CHARCONV_ENDIAN_LITTLE_BYTE != 0))
```
mborland, line 115:
```suggestion
#if (defined(BOOST_DECIMAL_ENDIAN_LITTLE_BYTE) && (BOOST_DECIMAL_ENDIAN_LITTLE_BYTE != 0))
```

include/boost/decimal/detail/emulated128.hpp
```diff
@@ -10,14 +10,15 @@
...
 #include <limits>
-#include <iosfwd>
+#if !defined(BOOST_DECIMAL_DISABLE_IOSTREAM)
 #include <iomanip>
+#include <iosfwd>
 #include <ostream>
-#include <cstdint>
-#include <cassert>
-#include <cmath>
+#endif
```
mborland, line 20: Move to end

mborland:
I think this looks good. I added https://github.com/cppalliance/decimal/issues/326 since we'll need to document all the new user configuration macros.

include/boost/decimal/detail/bit_layouts.hpp
ckormanyos: In the terse form, if it is not defined at all, then checking its value only leads to warnings (properly so) with certain warnings activated.

include/boost/decimal/detail/apply_sign.hpp
ckormanyos: This is a subtle one. I believe since fundamental types such as `int`, `short` or `long` _might_ differ on `arm-none-eabi-gcc` than they are from other compilers, that the construction from `int128` is ambiguous. The compiler issues not a warning fut a hard error. I have not figured out the exact place to correct these. Probably a whole bunch of `enable_if`s on the ctors of the emulated 128-bit integer.

See image here.

grafik

ckormanyos: We need to correct something or this will creep up on other compilers that have maybe even wider (nit lss-wide) types as well. So I need to stick with some kind of cast since these changes are not in the scope of the metal PR.

include/boost/decimal/detail/bit_layouts.hpp
ckormanyos: In the terse form, if `OOST_CHARCONV_ENDIAN_LITTLE_BYTE` is not defined at all, then checking its value only leads to warnings (properly so) with certain warnings activated.

ckormanyos:
Thanks Matt (@mborland) I've addressed your review comments. Some do, however, remain with work to do for instance on `int128` but that cna't be done in this PR.

include/boost/decimal/detail/bit_layouts.hpp
mborland: I meant it as that's a copy paste error. The macro should be BOOST_DECIMAL not BOOST_CHARCONV

include/boost/decimal/detail/apply_sign.hpp
mborland: That's annoying because all the integer constructors of int128 are fixed width types on purpose: https://github.com/cppalliance/decimal/blob/develop/include/boost/decimal/detail/emulated128.hpp#L600. I think you can drop the `INT8_C` since now it's being casted to int8_t twice.

ckormanyos:
Hi Matt (@mborland) it is all cycling now... Let's see

ckormanyos:
> The benchmark calculates the approximate value:

New data are in after Knuth long division, to be compared with the table in the [post above ](https://github.com/cppalliance/decimal/pull/316#issuecomment-1809722405).

Compare the final line(s) in the table(s). The benchmark on the metal is now about $6$-times faster and we save about $50\\%$ code size (probably has other reasons).

The bare metal benchmark on ARM(R) Cortex(R) M4 computes:

$$
e^{1.234}{\sim}3.43494186080075997{\ldots}
$$

| Type                           |  runtime [us] |  relative  |    code-size [kb] |
|--------------------------------|---------------|------------|-------------------|                               
| `double` (built-in, no FPU)      |  22           |   1.0      |       5.6         |
| `::math::softfloat::float64_t`   |  27           |   1.2      |       8.5         |
| `boost::decimal::decimal64`      |  490         |   22      |       20          |

Cc: @mborland

PR #331, opened 2023-11-17
Add cmath friend functions

mborland:


mborland:
@ckormanyos I am thinking hard about dropping GCC-6 since the failures keep racking up, and they are no longer contained to just the 32-bit address mode. Since every other compiler gets it right, and they are computational issues there must be some problem with code generation. What do you think?

Issue #341, opened 2023-11-21
Overloads for std::format


mborland:
https://www.cppstories.com/2022/custom-stdformat-cpp20/

PR #363, opened 2023-12-13
Support for <format>

mborland:
Closes: #341 

I am going to leave this one as a draft for now. There are a number of compiler errors since I don't think the support for this is totally complete yet.


mborland:
Use: https://github.com/boostorg/multiprecision/pull/535/files as reference for implementation

mborland:
GCC-13 and Clang-17 added full support for `<format>` so this should be feasible now.

PR #373, opened 2024-01-09
Enable coverage of tests

mborland:


ckormanyos (commented):
I wish there were a better way to do this kind of thing, ... I have not yet found one other than line-by-line, file-by-file. I guess it makes sense for exceptions. It's just tough when like you are doing correcting for a technical debt. Then you get lots of editing. Cool operation in this PR, Matt (@mborland)

mborland:
> I wish there were a better way to do this kind of thing, ... I have not yet found one other than line-by-line, file-by-file. I guess it makes sense for exceptions. It's just tough when like you are doing correcting for a technical debt. Then you get lots of editing. Cool operation in this PR, Matt (@mborland)

It's a little mind numbing but should be good now. I found at lease one test that was being completely skipped by too narrow constraints. I found that a lot of the boost libraries cover the test directory, and it seems like good practice.

ckormanyos:
> should be good now

Indeed very good Great stuff @mborland

Issue #377, opened 2024-01-12
asin for decimal128


mborland:
Currently coefficients only exist for 32 and 64 bit types

PR #378, opened 2024-01-12
Asin support for decimal128

mborland:
Closes: #377

### DIFF
--- a/include/boost/decimal/detail/cmath/impl/asin_impl.hpp
+++ b/include/boost/decimal/detail/cmath/impl/asin_impl.hpp
@@ -6,6 +6,8 @@
 #define BOOST_DECIMAL_DETAIL_CMATH_IMPL_ASIN_IMPL_HPP
 
 #include <boost/decimal/detail/concepts.hpp>
+#include <boost/decimal/detail/emulated128.hpp>
+#include <boost/decimal/detail/cmath/fma.hpp>
 #include <cstdint>
 
 namespace boost {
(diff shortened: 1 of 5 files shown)


include/boost/decimal/detail/cmath/impl/asin_impl.hpp
```diff
@@ -14,62 +14,81 @@ namespace boost {
...
+                                                                     decimal64 {UINT64_C(7499992533825458566), -20},
+                                                                     decimal64 {UINT64_C(2034140780525051207), -27},
+                                                                     decimal64 {UINT64_C(1666666666327808185), -19},
+                                                                     decimal64 {UINT64_C(2987315928933390856), -31},
+                                                                     decimal64 {UINT64_C(9999999999999989542), -19},
+                                                                     decimal64 {UINT64_C(6087279793251991118), -37}
+                                                             }};
+};
+
+#if !(defined(__cpp_inline_variables) && __cpp_inline_variables >= 201606L) && (!defined(_MSC_VER) || _MSC_VER != 1900)
```
ckormanyos, line 75: Hi Matt, could we turn the logic into an intuitive PP symbol?

```
#if !(defined(__cpp_inline_variables) && __cpp_inline_variables >= 201606L) && (!defined(_MSC_VER) || _MSC_VER != 1900)
```

mborland: I'll make it `BOOST_DECIMAL_HAS_INLINE_VARIABLES`. Without this you'll get all kinds of linker errors.

include/boost/decimal/detail/cmath/impl/asin_impl.hpp
```diff
@@ -14,62 +14,81 @@ namespace boost {
...
-    decimal64 {UINT64_C(2987315928933390856), -31},
-    decimal64 {UINT64_C(9999999999999989542), -19},
-    decimal64 {UINT64_C(6087279793251991118), -37}
-}};
+// Use a struct of arrays so that we can have static constexpr arrays of coefficients.
+// If we don't do this we will end up constructing the values in the arrays each time the function is called
+// which leads to massive slowdowns.
+//
+// See https://github.com/boostorg/math/issues/923 for further information
+template <bool b>
```
ckormanyos, line 24: Could we use a more unique and unequivocal name for the dummy Boolean parameter instead of `b`? Maybe use something like `BooleanDummyParam`.

include/boost/decimal/detail/cmath/impl/asin_impl.hpp
ckormanyos: Awesome. Nice choice and many thanks.

PR #379, opened 2024-01-14
Start preliminary lgamma skeleton

ckormanyos:


mborland:
> Hi Matt (@mborland) I finally had a moment to implement the skeleton for `decimal`'s elementary function `lgamma`. But CI is failing in a seemingly unrelated area. I'm not sure how to handle this. Maybe you could take a look?

All the failures are from `lgamma` that I see. Looks to be a slightly too tight tolerance? Most of them seem to be on 32-bit runs too.

include/boost/decimal/detail/cmath/lgamma.hpp
```diff
@@ -0,0 +1,230 @@
...
+#include <boost/decimal/detail/type_traits.hpp>
+#include <boost/decimal/detail/concepts.hpp>
+#include <boost/decimal/numbers.hpp>
+
+namespace boost {
+namespace decimal {
+
+// Begin: Forward declarations
+template <BOOST_DECIMAL_DECIMAL_FLOATING_TYPE T>
+constexpr auto log(T x) noexcept -> std::enable_if_t<detail::is_decimal_floating_point_v<T>, T>;
```
mborland, line 22: Do you need these forward declarations? You should be able to just include the relevant headers since you won't end up with a cycle.

include/boost/decimal/detail/cmath/lgamma.hpp
```diff
@@ -0,0 +1,230 @@
...
+                    #if (defined(__clang__) && (__clang__ < 6))
+                    #  pragma clang diagnostic push
+                    #  pragma clang diagnostic ignored "-Wmissing-braces"
+                    #endif
+
+                    constexpr auto coefficient_table =
+                        coefficient_array_type
+                        {
+                            // Use a Taylor series expansion of the logarithm of the gamma function.
+                            // N[Series[Log[Gamma[x]], {x, 0, 26}], 32]
```
mborland, line 95: Thanks for including the Wolfram command. I can generate the decimal128 constants once this is merged.

include/boost/decimal/detail/cmath/lgamma.hpp
ckormanyos: > Do you need these forward declarations?

I suppose they are not needed.

ckormanyos:
>  Looks to be a slightly too tight tolerance?

Yes there is a tolerance issue also.

But I also see something about not being able to convert decimal32 to double. It is in what seems to be an unrelated test case. Are there maybe multiple things going on?

mborland:
> >  Looks to be a slightly too tight tolerance?
> 
> 
> 
> Yes there is a tolerance issue also.
> 
> 
> 
> But I also see something about not being able to convert decimal32 to double. It is in what seems to be an unrelated test case. Are there maybe multiple things going on?
> 
> 
> 
> 

Oh there's a test case for concepts that spits out a million errors but it's marked as expected to fail in the jam file. If you search for ...failed you'll find the test cases that actually failed.

ckormanyos:
> find the test cases that actually failed

Ugghhh. Some of these (maybe all of them) are `lgamma(x)` for $x{\sim}1$ or $x{\sim}2$, for which the result is approximately equal to the logarithm of $1$. We lose bits here in subtraction. I have not handled the special cases near $1$ or $2$ yet. So I'll increase both the tolreances as well as the widths of the windows avoiding these arguments.

I'm not sure what's best to do with these points as we progress, maybe for this project, just leave them open and imprecise.

Changes forthcoming

ckormanyos:
Hi Matt (@mborland) this is looking rather green now. Got a bunch of pretty solid tests for `lgamma` included now.

I'm not quite sure how to handle two-time failure of CodeCov to upload? Wait? Try again? Force push?

mborland:
> I'm not quite sure how to handle two-time failure of CodeCov to upload? Wait? Try again? Force push?

Since it's internal server error I would just merge once the actual CI tests are green and come back to it in a day or so.

ckormanyos:
Hi Matt (@mborland) `lgamma` is finished for 32/64 in this PR. CodeCov is still 502-ing. I do not have the access to force-push. I'm happy to wait for CodeCove to come to its senses. or leave it in your hands.

Whatever you diecide, the technical part of `lgamma` impl is looking good and green for this early stage of the project.

We can also kick this one down the road a day or two to see what's up with CodeCov. Your choice.

ckormanyos:
Looks like CodeCov might be Baaaack. Let's see...

mborland:
> > CodeCov is still 502-ing.
> 
> 
> 
> Looks like CodeCov might be Baaaack. Let's see...

Everything is green now.

PR #380, opened 2024-01-15
Template coefficient lists for tables

ckormanyos:


ckormanyos:
Hi Matt (@mborland) as discussed in an offline chat, `tgamma` shows how we _might_ specializew coef-lists. Let's see if the compilers _like_ this stuff?

include/boost/decimal/detail/cmath/tgamma.hpp
```diff
@@ -17,6 +17,62 @@
 namespace boost {
 namespace decimal {
 
+namespace detail {
+
+template <BOOST_DECIMAL_DECIMAL_FLOATING_TYPE T,
+          const std::size_t CoefsSize>
+std::array<T, CoefsSize> coefs_array_table = {{ T { } }};
```
mborland, line 24: Should this be `static constexpr`?

ckormanyos: > Should this be `static` `constexpr`

Yes. I just saw that on the secong go-around.

An interesting thing did, however, happen when generating a list of $44$ coefficients for the $128$-bit case. The number of `constexpr` iterations exceeded the default allowed value of around a million for some compilers. Now all these compilers have ways to set the maximum number of allowed `constexpr` iterations, but I usually draw the line at the million.

What caught my eye was _where_ the iterations went. lots of shifting and I'm startnig to really want more and more to rework long division for big ints in this project...

ckormanyos:
Matt (@mborland) on both my recent merge to develop as well as in this PR, one particular CI runner is stopping, repeatedly, now twice in a row, with the message:

```
##[error]The operation was canceled.
```

Do you know what might be going on? It's nowhere near the 5 hour limit. Is there a systematic problem I may have introduced with `lgamma`? Or something just not well-understood?

mborland:
> Matt (@mborland) on both my recent merge to develop as well as in this PR, one particular CI runner is stopping, repeatedly, now twice in a row, with the message:
> 
> ```
> ##[error]The operation was canceled.
> ```
> 
> Do you know what might be going on? It's nowhere near the 5 hour limit. Is there a systematic problem I may have introduced with `lgamma`? Or something just not well-understood?

Timeout is set to 90 mins right now. You can bump the `timeout-limits` line from 90 to 120 or whatever.

ckormanyos:
> Timeout is set to 90 mins right now.

Ah OK. I went for `timeout-minutes: 240`, which is an hor less than the GHA maximum of 5 hours.

Checkin handling review comments soon.

ckormanyos:
> Should this be `static` `constexpr`

Aaarrghhh, my embedded compiler says:

```
../../include/boost/decimal/detail/cmath/tgamma.hpp:27:1: error: explicit template specialization cannot have a storage class
```

Oh no, who is right/wrong? Does anyone really comprehend what the ODR says in this case?

ckormanyos:
Actually basically all non-MSVC compilers told me to remove `static` from the explicit template specializations.

mborland:
> Actually basically all non-MSVC compilers told me to remove `static` from the explicit template specializations.

Without it being static I think we are going to plow into something similar to this issue: https://github.com/boostorg/math/issues/923. You'll end up constructing every one of those values every time which will kill performance.

ckormanyos:
> Without it being static I think we are going to plow into something similar to this issue: [boostorg/math/issues/923](https://github.com/boostorg/math/issues/923).

Ah yes I vaguely remember that nightmare. It was also iin the dead of winter... But last winter.

My compilers won't let me use the work static in some cases. Look at some of the logs from the just-cancelled CI run. Do you know which compilers do/do-not need `static` or which Boost keyword is available for this kind of thing?

mborland:
> > Without it being static I think we are going to plow into something similar to this issue: [boostorg/math/issues/923](https://github.com/boostorg/math/issues/923).
> 
> Ah yes I vaguely remember that nightmare. It was also iin the dead of winter... But last winter.
> 
> My compilers won't let me use the work static in some cases. Look at some of the logs from the just-cancelled CI run. Do you know which compilers do/do-not need `static` or which Boost keyword is available for this kind of thing?

The solution John used was to make a struct of arrays. I am going to try that now in the asin PR.

ckormanyos:
> The solution John used was to make a struct of arrays. I am going to try that now in the asin PR.

Ugghhh what a mess. I don't know how to deal with that either because I thought we'd get problems with the aggregate initialization (since that was C++17). I'll just wait and see what syntax you select in the `asin` PR.

ckormanyos:
Running into a tolerance issue anyway (I got greedy and tried to reduce the tolerances, but that was a mistake).

ckormanyos:
Hi @mborland I've immitated (but not exaclty) your method from the `asin` PR. Could you please tell me what you think of the detail-impl of the `tgamma` coefficients now? I think I paired it down to the essentials, but I'd like a second opinion.

include/boost/decimal/detail/cmath/impl/tgamma_impl.hpp
```diff
@@ -0,0 +1,100 @@
...
+        +::boost::decimal::decimal64 { UINT64_C(1'226'778'628'238'260'790), - 19 - 14 }, // * z^25
+        -::boost::decimal::decimal64 { UINT64_C(1'181'259'301'697'458'770), - 19 - 15 }, // * z^26
+        +::boost::decimal::decimal64 { UINT64_C(1'186'692'254'751'600'333), - 19 - 17 }, // * z^27
+    }};
+};
+
+} //namespace tgamma_detail
+
+template<typename T>
+using tgamma_table_struct_type = detail::tgamma_detail::tgamma_table_imp<T>;
```
mborland, line 94: I do like this a little better. You'll still need the macro for inline variables otherwise on some systems you'll get linker errors from the static table inside of the struct.

ckormanyos:
> need the macro for inline variables otherwise on some systems you'll get linker errors from the static table 

Wow. Indeed, I'm getting a big giant red-wave of them in CI...

ckormanyos:
Hi @mborland I think we need to update the required jobs in branch protection in pics below.

grafik

grafik

PR #387, opened 2024-01-17
Wide int

ckormanyos:
The purpose of this PR is to simply use wide-integer's Knuth long-division algorithm unchanged from the original library. At first it may be quite a mash-up. If this gives a significant performance gain and/or reduction of `constexpr` iterations, then we can discuss ways to select _only_ what is actually needed from wide-integer.


ckormanyos:
In change set 25d9b8110dee3309d14dde2e4b0fa7c91382eb3a, only the 128-bit division has been modified. It directly uses wide-integer to perform long division.

It is a lot of header inclusion. Let's see if the compilers _like_ this prior to going all in for 256-bit as well.

ckormanyos:
Hi @mborland cool promis.

I did a run of some randomly chosen non-trivial `decimal32` elementary funcs locally both with/without wide-integer`s division. The change is profound. See pics.

DEVELOP (branch)
grafik

WIDE_INT (branch)
grafik

ckormanyos:
I can't wait to see if all of CI passes the 128-bit case and how this plays out for int 256 and beyond if needed for `decimal64` and `decimal128`, ...

mborland:
> Hi @mborland cool promis.
> 
> I did a run of some randomly chosen non-trivial `decimal32` elementary funcs locally both with/without wide-integer`s division. The change is profound. See pics.

That is a pretty impressive speedup. I guarantee you the 256-bit changes will be drastic. Codecov is probably going to have a fit at first since you're using such a small percentage of that file.

ckormanyos:
> the 256-bit changes will be drastic.

I'm waiting to see if/when 128-bit actually passes all of CI.

> Codecov is probably going to have a fit at first since you're using such a small percentage of that file

At the end we will only need Knuth long division and one single linear-order multiplication of $n{\times}1$. We can pair it down to these routines only and use CodeCov as our guide to what we need/don't need.

As a next step, just then re-write the Knuth long division to be better compatible with `decimal`'s interpretation of bigint's. Prefer evolution to revolution.

ckormanyos:
Some of the older compilers are not comprehending the `constexpr`-ness of wide-integer. I think it might take me a couple of rounds to track it down...

ckormanyos:
Trying for 256-bit long division now.

Although there are some systematic failuers in `test_atan2.cpp` for two older compiler runs, I'm plowing ahead for 256-bit, assuming the easy stuff can be sorted out ASAP.

Very, very, very preliminary performance check:

DEVELOP branch:
grafik

WIDE-INT branch:
grafik

ckormanyos:
Hi Matt (@mborland) long division is looking promising. All is fast by factors of $2{\ldots}100}$.

MSVC CI runs now in an hour.

There is, however, one pesky problem in `test_atan2.cpp` on old compilers. I suspect it's bogus, but U nvr know.

Finish this off maybe in a few days. If you have keen insight into the few failures on `test_atan2.cpp`, please let me know. Otherwise we hunt them down together.

Yeah getting fast(er).

ckormanyos:
Down the road, I'll simplify this to the bare-bones, what decimal needs. Get down to a cople hundred lines. But it's looking promising for the moment.

mborland:
> Hi Matt (@mborland) long division is looking promising. All is fast by factors of $2{\ldots}100$.
> 
> 
> 
> MSVC CI runs now in an hour.
> 
> 
> 
> There is, however, one pesky problem in `test_atan2.cpp` on old compilers. I suspect it's bogus, but U nvr know.
> 
> 
> 
> Finish this off maybe in a few days. If you have keen insight into the few failures on `test_atan2.cpp`, please let me know. Otherwise we hunt them down together.
> 
> 
> 
> Yeah getting fast(er).
> 
> 

That's a pretty impressive speed up. Thanks for this. I should have time to look through tomorrow after addressing the increasing amount of review issues in charconv.

ckormanyos:
> That's a pretty impressive speed up. Thanks for this. I should have time to look through tomorrow

Word.

There is no hurry on this. I simply had a _rare_ block of time to hammer this in. I'm surprised how well it went. Compliments to your excellent design.

We can finish this off in a while. But one thing is sure, you need this stuff. You go from like 100,000 iterations to maybe 200. in the `decimal-128`-bit case.

ckormanyos:
Hi Matt we are down to two unusually-failing test cases on selected compilers only in `test_atan2.cpp`.

Tests are cycling now which print rudimentary debug info (hex-dump) at the failing test cases. I'll look for the text _Chris hex-dump_ and maybe gain more insight into what's going on with these particular compiler(s).

ckormanyos:
OK. Based on ideas from an offline chat, add the _number_ `pi_over_four_v<>` and use it consistently. And, ... green we go.

This needs to cycle a few times for certainty and (coming up) to restore all pedantic warnings on as errors in CI. But that'll wrap up today.

I'll leave a few points open such as:
- Optimize even more.
- Reduce (even further) the scope of introduced wide-integer code.

Cc: @mborland

ckormanyos:
See also #388

ckormanyos:
CodeCov reveals. Before leaving this PR, I need to reliably hit [line 1168](https://app.codecov.io/gh/cppalliance/decimal/pull/387/blob/include/boost/decimal/detail/wide-integer/uintwide_t.hpp#L1168).

ckormanyos:
> Before leaving this PR, I need to reliably hit, ...

I left this PR as finished, anyway. And I will address this and other related issues in #388.

Issue #428, opened 2024-03-05
tgamma and lgamma use `std::crbegin()` which is not constexpr for C++14


mborland:
Something like: https://github.com/cppalliance/decimal/blob/develop/include/boost/decimal/detail/cmath/impl/remez_series_result.hpp should be used to replace the current implementations

PR #432, opened 2024-03-05
Move gamma coefficients into structs and fix 428

mborland:
Closes: #428 

CC: @ckormanyos I was able to get this working just fine. `static_asserts` have been added for tgamma and lgamma for decimal128 usage. I think the coefficient lists for decimal32 could be reduced since they are the same as dec64, but you're the expert on this one.


ckormanyos:
Hi Matt (@mborland), I really can't get the 128-bit series working. I tried for a few hours and gave up. I don't know if the intent is that 128-bit `tgamma()`should or should not work. But I keep getting the wrong results at 128-bit using code that goes a little beyond your commit.

mborland:
> Maybe we are in different areas, but I'm explicitly going for 128-bit `tgamma()`. Or is this out of scope at the moment?

Go for it. If you push the branch I can look at it tomorrow if need be as well.

ckormanyos:
Hi @mborland indeed i ran into confusion. `tgamma()`-jama.

In this trivial push, I just ry to use the existing coefs for 128-bit. But then get the wrong answer for:

$${\Gamma}\dfrac{456}{100}$$

```
#include <boost/decimal.hpp>

#include <iomanip>
#include <iostream>
#include <limits>
#include <sstream>

auto main() -> int
{
  using local_decimal_type = boost::decimal::decimal128;

  const local_decimal_type x = local_decimal_type { 456, -2 };

  const auto tg = tgamma(x);

  {
    std::stringstream strm;

    strm << std::setprecision(std::numeric_limits<local_decimal_type>::digits10) << x;

    std::cout << strm.str() << std::endl;
  }

  {
    std::stringstream strm;

    strm << std::setprecision(std::numeric_limits<local_decimal_type>::digits10) << tg;

    std::cout << strm.str() << std::endl;
  }
}
```

Issue #440, opened 2024-03-06
cbrt(8) wrong result


ser90368141:
Code for test:
```
using decimalType = boost::decimal::decimal64;
std::cout << std::setprecision(std::numeric_limits<decimalType>::digits10 +
                               1)
          << "pow(8, 1/3) = "
          << boost::decimal::pow(decimalType{8}, decimalType{1} / 3)
          << std::endl;
std::cout << std::setprecision(std::numeric_limits<decimalType>::digits10 +
                               1)
          << "cbrt(8) = " << boost::decimal::cbrt(decimalType{8})
          << std::endl;
```
Output:
```
pow(8, 1/3) = 1.999999999999996e+00
cbrt(8) = 8.326831776556042e-01
```

ser90368141:
Also
```
std::cout << std::setprecision(std::numeric_limits<decimalType>::digits10 +
                               1)
          << "pow(-8, 1/3) = "
          << boost::decimal::pow(decimalType{-8}, decimalType{1} / 3)
          << std::endl;
```
`pow(-8, 1/3) = nan`

mborland:
@ckormanyos When you get a chance can you please take a look at this one. I thought my newton steps were correct, but clearly something has gone horribly wrong.

ckormanyos:
> When you get a chance can you please take a look at this one.

Yes I’ll be happy too. Some of these so-called _easy_ functions can turn out to be remarkably difficult.

I’ve also still got a TODO on the poor performance of our square root.

mill look into this bug ASAP

PR #441, opened 2024-03-07
Fix cbrt and expand testing

mborland:
Fixes: #440 

@ckormanyos I simplified the step and initial guess which has this working correctly. I still think it's worth investigating a much better initial guess if that's something you still want to pursue. 

I am going to keep separating tests out of the giant test_cmath file because I found this was tested incorrectly (wasn't taking the abs of the float distance) and it's easier to run GDB through smaller files.


ckormanyos:
> I still think it's worth investigating a much better initial guess if that's something you still want to pursue.

Yes @mborland I am interested but simply excessively booket out. Bosh this issue as well as the poor initial guess on square root are on my TODO list.

PR #521, opened 2024-05-01
T gamma bama jama

ckormanyos:
Finally PR-ing `tgamma()` at $128$ bits. With one test and a few TODOs. But let's see how CI cycles...

This PR is also interested in the bug fix for #520


ckormanyos:
Hi Matt (@mborland) `tgamma()` at 128 bit is looking good. I added a few more tests and its cycling now. If I end up merging this, then we need to replace my homegrown impls of `fma()` with the real one as soon as #520 gets fixed as well.

ckormanyos:
Hi Matt (@mborland) I'm going to go ahead and merge this one to get the ball rolling again. There are some local impls of `fma()` that I'll remove when develop has the 128-bit `fma()` fix.

PR #526, opened 2024-05-03
Faster zero removal

mborland:
@jk-jeon I thought you might be interested in these results. I took the branchless Granlund-Montgomery from your [repo](https://github.com/jk-jeon/rtz_benchmark/blob/master/source/main.cpp) for `std::uint32_t` and `std::uint64_t` and then adapted a 128-bit case. I also added a test set for every representable power of 10 to see if the result is always 1. I found that in the 32 bit and 64 bit cases an extra step was needed beyond what you had in your repo as it was stopping at 7 and 15 zeros removed respectively. In the 128-bit case I ended up going for two passes since with t = 32, there is no modular inverse so the whole calculation becomes garbage.

### DIFF
--- a/include/boost/decimal/detail/remove_trailing_zeros.hpp
+++ b/include/boost/decimal/detail/remove_trailing_zeros.hpp
@@ -95,14 +95,17 @@ constexpr auto remove_trailing_zeros(std::uint64_t n) noexcept -> remove_trailin
     return {n, s};
 }
 
-namespace impl {
-
-constexpr auto remove_trailing_zeros_impl(uint128 n) noexcept -> remove_trailing_zeros_return<uint128>
+constexpr auto remove_trailing_zeros(uint128 n) noexcept -> remove_trailing_zeros_return<uint128>
 {
     std::size_t s {};
 
-    auto r = rotr<128>(n * uint128 {UINT64_C(0x3275305C1066), UINT64_C(0xE4A4D1417CD9A041)}, 16);
-    auto b = r < uint128 {UINT64_C(0x734), UINT64_C(0xACA5F6226F0ADA62)};
+    auto r = rotr<128>(n * uint128(UINT64_C(0x62B42691AD836EB1), UINT64_C(0x16590F420A835081)), 32);
+    auto b = r < uint128 {UINT64_C(0x0), UINT64_C(0x33EC48)};
+    s = s * 2U + static_cast<std::size_t>(b);
+    n = b ? r : n;
+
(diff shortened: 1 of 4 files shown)


jk-jeon:
Hi, thanks for letting me know!

In my intended specific application, 32-bit numbers have at most 8 digits and 64-bit numbers have at most 16 digits, so I intentionally omitted the extra step. Of course `t=8` step and `t=16` step are needed if there is no a priori known upper bounds.

For the 128-bit case, I am a bit confused. So it seems the largest power of 10 fitting in 128-bits is $10^{38}$, and $5^{32}$ should have its modular inverse with respect to $2^{128}$. According to [Wolfram alpha](https://www.wolframalpha.com/input?i=hex%28PowerMod%5B5%5E32%2C+-1%2C+2%5E128%5D%29), it's `0x62b42691'ad836eb1'16590f42'0a835081`.

Here is a test code: https://godbolt.org/z/WMrvWhb41

The generated assembly is not particularly adorable, but I guess unless you can ensure an upper bound on the input (better than $2^{128}-1$) this is probably the best I can do.

mborland:
> For the 128-bit case, I am a bit confused. So it seems the largest power of 10 fitting in 128-bits is 1038, and 532 should have its modular inverse with respect to 2128. According to [Wolfram alpha](https://www.wolframalpha.com/input?i=hex%28PowerMod%5B5%5E32%2C+-1%2C+2%5E128%5D%29), it's `0x62b42691'ad836eb1'16590f42'0a835081`.
> 
> Here is a test code: https://godbolt.org/z/WMrvWhb41
> 

Thanks for that. I'm not sure why I was getting m = 0 but this is clearly correct.

> The generated assembly is not particularly adorable, but I guess unless you can ensure an upper bound on the input (better than 2128−1) this is probably the best I can do.

Unfortunately here the upper bound on the input is digits10 - 1 which does nothing to help.

jk-jeon:
>Thanks for that. I'm not sure why I was getting m = 0 but this is clearly correct.

I observed that according to my post the multiplier is supposed to be the modular inverse with respect to $2^{128-32}$ rather than $2^{128}$ so maybe that's a source of confusion. It doesn't matter in the end though. It seems you followed that convention, which resulted different multipliers from what I did in the godbolt link. For consistency, I guess it's probably a little bit better to continue that convention which means to remove the first 8 hex digits from the multiplier for `t=32` step. Nothing changes in terms of functionality though.

>Unfortunately here the upper bound on the input is digits10 - 1 which does nothing to help.

I assume you mean the bound is $10^{38}-1$. That little bit of margin allowed me to try the generalized modular inverse method which replaces `rotr` by the regular shift, but it seems to generate much worse assembly: https://godbolt.org/z/b9Wf463vP
So the status quo is the best I guess.

PR #527, opened 2024-05-04
Impl of cosh/exp at 128 bits and a few tests

ckormanyos:


ckormanyos:
This is intended to handle #522

ckormanyos:
Hi Matt (@mborland) I really am enjoying the `decimal` experience. Really a lot of the early troubles and woes are gone. It's really getting fun to work with. Crisp. Clean.

While handling this PR, I gazed at the runtime of CI up over an hour. I think we have more VMs than time nowdays (maybe also at boost) but for sure at the alliance. Along these lines, do you think it makes sense to split CI up into smaller constituent parts? I can't really figure out the right mix. I think on that level, we would have to split up the Jamfile since we have all these VMs whirling away all the time.

Do you know if there is a model to split up your main Jamfile? Is this interesting to you/us?

ckormanyos:
> a model to split up your main Jamfile?

Hi Matt (@mborland) or split up the YAML into two almost identical YAML files where the only difference is that the test matrices (language standards or better yet old/new compilers) are split between two or among several almost-identical YAML files. Thereby leave Jamfile alone.

Thoughts?

mborland:
> Hi Matt (@mborland) I really am enjoying the `decimal` experience. Really a lot of the early troubles and woes are gone. It's really getting fun to work with. Crisp. Clean.
> 
> 
> 
> While handling this PR, I gazed at the runtime of CI up over an hour. I think we have more VMs than time nowdays (maybe also at boost) but for sure at the alliance. Along these lines, do you think it makes sense to split CI up into smaller constituent parts? I can't really figure out the right mix. I think on that level, we would have to split up the Jamfile since we have all these VMs whirling away all the time.
> 
> 
> 
> Do you know if there is a model to split up your main Jamfile? Is this interesting to you/us?

I think our timeout limit is 3 hours so we should be ok for now. As we dive into more special functions for 128-bit we could look at splitting up more or reducing the number the number of values tested.

PR #536, opened 2024-05-05
Logarithmic funcs

ckormanyos:


ckormanyos:
This PR is intended to address #534 and #535

include/boost/decimal/detail/cmath/log.hpp
```diff
@@ -164,29 +93,6 @@ constexpr auto log_impl(T x) noexcept
...
-    using evaluation_type = detail::promote_args_t<T, decimal64>;
-
-    #else // BOOST_DECIMAL_DEC_EVAL_METHOD == 2
-
-    using evaluation_type = detail::promote_args_t<T, decimal128>;
-
-    #endif
-
-    return static_cast<T>(detail::log_impl(static_cast<evaluation_type>(x)));
-}
```
mborland, line 188: We need to keep this part on each of the special functions to properly support the float evaluation methods. This promotes the type of x depending on the macro set and then does the calculation much like how boost math does.

PR #540, opened 2024-05-07
Restore mistakenly removed promotions

ckormanyos:


ckormanyos:
Hi Matt (@mborland) somewhere I think I forgot or messed up a promotion. It _smells_ like `atanh()`. But I don't recall modifying that. Or I just plain old forgot.

Can you see what's bugging an old GCC on drone? I looked and can't figure it out...

mborland:
@sdarwin here's a new instance of GCC 7 hanging. Good news is it looks like the build is being killed now instead of running for unlimited time.

PR #542, opened 2024-05-07
Drone: divide GCC 7 jobs

sdarwin:


ckormanyos:
Thanks Sam (@sdarwin). Hi Matt (@mborland) I think CodeCov can't run for another contributor due to some token reason. I've often struggled with this kind of problem but never solved it.

mborland:
> Thanks Sam (@sdarwin). Hi Matt (@mborland) I think CodeCov can't run for another contributor due to some token reason. I've often struggled with this kind of problem but never solved it.

I can look into it; likley some security setting because Sam's branch is in a fork.

PR #544, opened 2024-05-09
Log10

ckormanyos:


include/boost/decimal/detail/cmath/log10.hpp
```diff
@@ -26,21 +27,102 @@ template <typename T>
...
+
+        int exp10val { };
+
+        T g { frexp10(x, &exp10val) };
+
+        while(g > one)
+        {
+            // TODO(ckormanyos) Do we really have to multiply individual powers
+            // of 10 here? Or is there a reliable way to simply use pow10()?
+            // For instance count intagral orders of exp10val and divide by pow10().
```
mborland, line 75: The significand is normalized before being returned by frexp10 so you can divide by pow10(digits10 - 1) to avoid the loop

ckormanyos: >  significand is normalized before being returned by frexp10 so you can divide by pow10(digits10 - 1) to avoid the loop

OK @mborland cool but now I'm stuck on another point. Do you know how to recognize a pure power of $10$ without burning too much CPU cycling?

mborland: You could use `detail::remove_trailing_zeros` and compare to 1. That is about as efficient as it's going to get I think.

ckormanyos:
> Got it. Will commit cool reaction to advice.

OK Matt (@mborland) I finally was able to implement a `log10()` with which I'm mostly satisfied. The newer stuff like trimming trailing zeros is very nice and convenient to work with.

I did notice, however, that there are several places in the code where we end up using triple-conditional type-definitions in order to arrive at `uint32_t`, `uint64_t` or the specialized 128-bit representaion type. Do you think it makes sense to collect these and support one single definition of the decimal-type's `representation_type`?

mborland:
> > Got it. Will commit cool reaction to advice.
> 
> 
> 
> OK Matt (@mborland) I finally was able to implement a `log10()` with which I'm mostly satisfied. The newer stuff like trimming trailing zeros is very nice and convenient to work with.
> 
> 
> 
> I did notice, however, that there are several places in the code where we end up using triple-conditional type-definitions in order to arrive at `uint32_t`, `uint64_t` or the specialized 128-bit representaion type. Do you think it makes sense to collect these and support one single definition of the decimal-type's `representation_type`?

I guess it would be easy to add the same typedef into each of the classes and just query that. I'm cool with it.

include/boost/decimal/detail/cmath/log10.hpp
```diff
@@ -26,21 +27,119 @@ template <typename T>
...
+        using representation_type = std::conditional_t<std::is_same<T, decimal32>::value, std::uint32_t,
+                                    std::conditional_t<std::is_same<T, decimal64>::value, std::uint64_t, detail::uint128>>;
+
+        const representation_type gn { frexp10(x, &exp10val) };
+
+        const remove_trailing_zeros_return<representation_type>
+            zeros_removal
+            {
+                remove_trailing_zeros(gn)
+            };
```
mborland, line 62: Couldn't all of this use of `representation_type` be replaced by `auto`? Then we wouldn't need to fuss with around with adding the typedef.

ckormanyos: > Couldn't all of this use of `representation_type` be replaced by `auto`?

Yes. Let me run that through lickety split.

PR #548, opened 2024-05-11
Impl expm1 64/128 and other small repairs

ckormanyos:


ckormanyos:
Intended to handle #537

PR #549, opened 2024-05-11
Preliminary step simplify atan()

ckormanyos:


ckormanyos:
Hi Matt (@mborland) I cleaned up `atan()` a bit. i took the liberty to use a well-known argument-scaling from the literature to reduce the number of coefficient lists from $4$ to $2$ (`small`, `med`, `large`, `huge`) to now only `small`, `med`).

I was not able to get the lists down to $1$. And I've given up on that.

I was not able to generate any well-working $128$-bit coefficients and simply stuck with the $64$-bit ones for $128$ also. So if we can generate two $128$-bit coefficient lists, the `atan()` will be easy to finish.

ckormanyos:
>  was not able to generate any well-working $128$-bit coefficients

I was able to reduce the needed sets of coefficients to one set. And I did, in fact, end up going with a Pade Approximation for small-arguments at $128$-bits. Is cycling now. I might reduce the number of parameters at 32 bits or just leave them.

Let's see how this shakes out in CI...

Issue #539, opened 2024-05-05
lgamma


ckormanyos:

ckormanyos:
While I'm at it, handle large-argument asymptotic expansion ot `tgamma()` in this issue as well.

ckormanyos:
Is cycling in #551

Issue #431, opened 2024-03-05
Add decimal128 coefficient tables to <cmath> functions


mborland:
- [x] #519
- [ ] #539
- [x] #538
- [x] #537
- [x] #533
- [x] #522
- [x] #528
- [x] #531
- [x] #534
- [x] #535

ckormanyos:
See also #381

Issue #381, opened 2024-01-15
Rework coef-lists 18 --> 19 decimal digits in elem funcs


ckormanyos:
... and tighten tols if possible or if this helps.

ckormanyos:
This pertains to 32/64 bit calcs.

ckormanyos:
This is parasitically being handled in #431

PR #551, opened 2024-05-13
Start finish gammas but lgamma-128 still needed

ckormanyos:


ckormanyos:
If/when this cycles, it is intended to handle #381, #431 and #539.

ckormanyos:
Almost forgot this one. But I did forget why it failed to cycle. Merge develop into it with the hope to close. One less issue. Let's let this one cycle through.

CC: @mborland

mborland:
Looks like a slight change of tolerance is needed in order to avoid spurious errors in the future.

ckormanyos:
OK I've added 128-bit for the gammas. It's not a perfect impl. But starting point good.

As we have lots more fundamental problems to address, I'll just adjust test tols and ranges and close the books (temporarily) on gammas until the higher priority tasks are dealt with.

Let's cycle this thingy.

mborland:
Can the tolerances go a bit looser? I'm seeing sporadic failures on CI: https://github.com/cppalliance/decimal/actions/runs/9157848785/job/25175026830#step:4:10586

PR #571, opened 2024-05-16
Update GCC13 runner os

mborland:
Github broke 22.04: https://github.com/actions/runner-images/issues/9866

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,13 +77,14 @@ jobs:
             os: ubuntu-22.04
             install:
               - g++-12-multilib
-          - toolset: gcc-13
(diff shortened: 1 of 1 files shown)


ckormanyos:
Hi Matt (@mborland) could you please look at the one failed run? It's not getting very far into it. Also a few comments above about `constexpr`-abs.

ckormanyos:
> the one failed run?

Ah I see, the runner is fundamentally flawed at the moment.

mborland:
Since we have GCC 13 covered in Drone I am just going to delete it out of GHA until the fix it

ckormanyos:
Why does stuff like this keep happenin' on such mighty runners???!!! Good choice Matt

Issue #559, opened 2024-05-15
uint256_t multiplication is incorrect


mborland:
@ckormanyos I am trying to improve the calculation for `num_digits` by pre-computing a binary search tree rather than doing repeated division for large types. The code for the generation is:

```
template <typename T, std::size_t N>
constexpr auto generate_array() noexcept -> std::array<T, N>
{
    std::array<T, N> values {};

    values[0] = T{1};
    for (std::size_t i {1}; i < N; ++i)
    {
        values[i] = values[i - 1] * UINT64_C(10);
    }

    return values;
}
```

For `uint256_t` the call is `constexpr auto powers_of_10 = generate_array<boost::decimal::detail::uint256_t, 79>();`. Using that tree I got a ton of failiures in basic operations. The debugger shows the below values in the array for example (i = 53, 54, ...):

<img width="353" alt="Screenshot 2024-05-15 at 9 09 40 AM" src="https://github.com/cppalliance/decimal/assets/3745830/83c9863a-62af-4290-845f-a98231225ee3">

Something seems to be going run on the high word.

ckormanyos:
Hi Matt (@ckormanyos). We talked in offline chats on this. And this might be what's causing "fuzz" around 30 decimal digits and higher in dec-128. . I'll try to get to the bottom of this one.

ckormanyos:
Hi Matt (@mborland) something is going haywire. I am tracking it down in a new branch. Give me a day or two to dial mul-256

Below is a re-posting of similar test code I adapted.

```cpp
#include <array>

#include <boost/decimal.hpp>

template <typename T,
          const std::size_t N>
constexpr auto generate_array() noexcept -> std::array<T, N>
{
  std::array<T, N> values { };

  std::size_t index { 0 };

  values[index] = T { 1 };

  ++index;

  for ( ; index < N; ++index)
  {
    values[index] = values[index - 1] * UINT64_C(10);
  }

  return values;
}

auto main() -> int
{
  using local_uint256_t = boost::decimal::detail::uint256_t;

  constexpr auto powers_of_10 = generate_array<local_uint256_t, static_cast<std::size_t>(UINT8_C(78))>();

  std::size_t idx { };

  for(const auto& ui_val : powers_of_10)
  {
    std::cout << "index: " << idx << ", " << ui_val << std::endl;

    ++idx;
  }
}
```

ckormanyos:
Now cycling in #574.

PR #574, opened 2024-05-18
Study mul256

ckormanyos:
The purpose of this PR is to simply insert parts of wide-integer's mul-256 into decimal. Let's see how CI shakes out.

### DIFF
--- a/include/boost/decimal/detail/emulated256.hpp
+++ b/include/boost/decimal/detail/emulated256.hpp
@@ -89,6 +89,8 @@ struct uint256_t
 
     friend constexpr uint256_t operator*(const uint256_t& lhs, const uint256_t& rhs) noexcept;
 
+    friend constexpr uint256_t operator*(const uint256_t& lhs, const std::uint64_t rhs) noexcept;
+
     friend constexpr uint256_t operator-(const uint256_t& lhs, const uint256_t& rhs) noexcept;
 
     constexpr uint256_t &operator-=(uint256_t v) noexcept;
@@ -208,37 +210,6 @@ constexpr uint256_t operator+(const uint256_t& lhs, const uint256_t& rhs) noexce
     }
 
     return temp;
-}
-
-constexpr uint256_t operator*(const uint256_t& lhs, const uint256_t& rhs) noexcept
-{
-    uint256_t result{};
-
-    uint128 products[4];
-
-    // Multiply the low parts.
-    uint128 product_low = static_cast<uint128>(lhs.low.low) * rhs.low.low;
-    products[0] = {uint64_t(product_low >> 64), uint64_t(product_low)};
-
-    // Multiply the mixed parts.
-    uint128 product_mid_low = static_cast<uint128>(lhs.low.high) * rhs.low.low;
-    uint128 product_mid_high = static_cast<uint128>(lhs.low.low) * rhs.low.high;
-    products[1] = {uint64_t((product_mid_low >> 64) + (product_mid_high >> 64)), uint64_t(product_mid_low + product_mid_high)};
-
-    // Multiply the high parts.
-    uint128 product_high = static_cast<uint128>(lhs.low.high) * rhs.low.high;
-    products[2] = {uint64_t(product_high >> 64), uint64_t(product_high)};
-
-    // Add the products, taking care of the carries.
-    bool carry = false;
-    result.low.add_with_carry(products[0], carry);
-    result.low.add_with_carry(uint128(products[1].low) << 64, carry);
-    result.high.add_with_carry(uint128(products[1].high, products[1].low) >> 64, carry);
-    result.high.add_with_carry(products[2], carry);
-
-    // Note: This does not handle overflow beyond 256 bits.
-
-    return result;
 }
 
 constexpr uint256_t operator+(uint256_t lhs, uint128 rhs) noexcept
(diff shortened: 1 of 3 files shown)


ckormanyos:
Hard error but very confusing on Ubuntu, `cbrt(128-bit)` for argument close to $1.0$ hangs, not fails but hangs indefinitely. Will now investigate why.

ckormanyos:
Hi Matt (@mborland) I took the opportunity to fully refactor `sqrt()` and `cbrt()` and their tests. Interesting point: When I had initially started with mul-256 for some strange reason the cube-root test was hanging. I also found that several tens of Newton iterations were being performed in some cases.

So for `sqrt()` and `cbrt()` I used a uniform estimation scheme providing slightly more than $2$ decimal digits of accuracy in the initial guess on the interval $1/10 {\leq} x < 1$. This results excellent results for exactly in $2$, $3$ or $4$ iterations for 32/64/128 bits respectively. I added and strengthened the tests and am now hammering the new multiplication and roots.

When implementing $8{\times}8$ multiplication, I opted to unroll the loop. I could add some specialized optimitations for multiplication $4{\times}4$ or $8{\times}2$, but we can do that later as needed.

ckormanyos:
Oh yes and also Matt (@mborland) there was, in fact, a hear error lurking around somewhere in the original $256{\times}256$-bit multiplication. So these changes (or something like them) are seemingly needed.

I can now reduce significantly all tolerances on `decimal128`-tests. I was tempted to gut out the whole thing and put in a small subset of wide-integer, but we've got no time for that at the moment. So I opted to simply port over a few routines.

Now the test codes in #559 all pass.

ckormanyos:
> ... can now reduce significantly all tolerances on `decimal128`-tests

See also #575.

ckormanyos:
Drone is having a bit of a tussle with some sort of scripts. Maybe we can sort this out. GHA is finally looking reasonable.

Hi Matt (@mborland) this PR has a bit of depth to it. Redo $256{\times}256$-bit mul, with future optimizations TODO. Also totally rework `sqrt()` and `cbrt()`.

Please give this thing a look at your convenience. I'm OK with it because $256$-bit math is now correct, or better-said, more correct. If there's any performance hits, we/I can probably deal with them. So I'd favor numerical correctness at this juncture.

include/boost/decimal/detail/cmath/cbrt.hpp
```diff
@@ -24,65 +24,149 @@ namespace decimal {
...
+            {
+                const int p10_mod3 = (p10 % 3);
+                const int p10_div3 = (p10 / 3);
+
+                result = T { 1, p10_div3 };
+
+                switch (p10_mod3)
+                {
+                    case 2:
+                        result *= numbers::cbrt10_v<T>;
```
mborland, line 81: `numbers::cbrt10_v` is new right? Can you please add it to the doc page on constants?

include/boost/decimal/detail/cmath/sqrt.hpp
```diff
@@ -24,63 +23,132 @@ namespace decimal {
...
+            }
+            else
+            {
+                const int p10_mod2 = (p10 % 2);
+
+                result = T { 1, p10 / 2 };
+
+                if (p10_mod2 == 1)
+                {
+                    result *= numbers::sqrt10_v<T>;
```
mborland, line 80: Same comment about adding to docs

include/boost/decimal/detail/emulated256.hpp
```diff
@@ -89,6 +89,8 @@ struct uint256_t
 
     friend constexpr uint256_t operator*(const uint256_t& lhs, const uint256_t& rhs) noexcept;
 
+    friend constexpr uint256_t operator*(const uint256_t& lhs, const std::uint64_t rhs) noexcept;
```
mborland, line 92: Do we also need `friend constexpr uint256_t operator*(const std::uint64_t lhs, const uint256_t& lhs) noexcept` for completeness?

include/boost/decimal/detail/emulated256.hpp
```diff
@@ -407,6 +378,58 @@ constexpr auto wide_integer_to_uint256(const wide_integer_uint256& src) -> uint2
...
+    using local_unsigned_fast_type = ::boost::decimal::math::wide_integer::detail::unsigned_fast_type;
+
+    // Mash-Up: Use unrolled school-multiplication from wide-integer (requires limb-conversions on input/output).
+
+    auto lhs_wide = uint256_to_wide_integer(lhs);
+    auto rhs_wide = uint256_to_wide_integer(rhs);
+
+    wide_integer_uint256 result_wide { };
+
+    wide_integer_uint256::eval_multiply_n_by_n_to_lo_part(result_wide.representation().begin(),
```
mborland, line 392: What's the type of `result_wide.representation()`? `std::array::begin` is not `constexpr` until C++17 and neither is `operator[]`.

test/test_lgamma.cpp
```diff
@@ -519,9 +530,7 @@ auto main() -> int
   }
 
   {
-    // TODO(ckormanyos) Can we reduce the tolerance on lgamma()-128?
-    // Can the approximation be done a bit better (Lanczos from Math?).
-    const auto result_lgamma128_is_ok   = local::test_lgamma_128(16'000'000);
+    const auto result_lgamma128_is_ok   = local::test_lgamma_128(8092);
```
mborland, line 533: Nice to see these tolerances went way down

include/boost/decimal/detail/emulated256.hpp
ckormanyos: > Do we also need `friend constexpr uint256_t operator* ...`

We need a bunch of them. And I could do a lot more optimizing for mixed-limb left/right. I would, however, prefer to get something into develop and subsequently optimize. Then I canwork on pressing down all the $128$-bit tolerances and other special functions, while optimizing when time arises.

test/test_lgamma.cpp
ckormanyos: Indee. See also #575

ckormanyos:
I was concerned about changing performance. Multiplication stays about the same. I also benchmarked `sqrt()` which has improved in the `mul256` branch compared to the `develop` branch. The tables are from MSVC 142, with similar trends on Ubuntu in WSL2.

| **Multiplication**           | `mul256` | `develop`|
|------------------------------|----------|----------|
|Multiplication `<float     >` | 443      | 558      |
|Multiplication `<double    >` | 476      | 609      |
|Multiplication `<decimal32 >` | 37383    | 37159    |
|Multiplication `<decimal64 >` | 125178   | 123256   |
|Multiplication `<decimal128>` | 397336   | 397468   |
|Multiplication `<dec32_fast>` | 32281    | 32995    |

| **`sqrt()`**       | `mul256` | `develop`|
|--------------------|----------|----------|
|`sqrt<float     >`  | 1053     | 1353     |
|`sqrt<double    >`  | 2908     | 2884     |
|`sqrt<decimal32 >`  | 417289   | 666491   |
|`sqrt<decimal64 >`  | 1604746  | 2181287  |
|`sqrt<decimal128>`  | 15742232 | 23100960 |

include/boost/decimal/detail/cmath/sqrt.hpp
ckormanyos: I added $\sqrt{10}$, $\sqrt[3]{10}$ and $\sqrt[3]{2}$. I can document these accordingly.

ckormanyos: Docs updated and slightly corrected. Please Matt (@mborland) could you make sure my corrections in the ASCII-docs are OK?

ckormanyos:
Hi Matt (@mborland) before leaving this one, I found another issue in `uint128` when performing $128{\times}64$-bit multiplication. I'll fix this here in this branch (although not with the highest possible efficiency).

At some time in the future, I'll take a few morning sessions and make a dedicated, C++14-friendly `uint`-whatever (like $128$, $256$) for this work. But we can probably push that off until the interfacing and most of the known numerical problems have been wliminated.

ckormanyos:
This thing fixes a bunch of pesky problems.

- Repair $128$ and $256$-bit multiplication, but retain whatever performance we have (repair temporary permanent fiy planned this year).
- Improve `sqrt()` and `cbrt()`.
- Eliminate spurious errors on standard function tests via eliminating troublesome ranges not needed for basic tests.
- Reduce `decimal128` tolerances on a bunch of tests.

PR #580, opened 2024-05-21
Start tuning cmath functions

ckormanyos:
Now that most `<cmath>` functions are present, the purpose of this PR is to tune them, check/extend ranges and try to hit more non-covered lines.


ckormanyos:
Hi Matt (@mborland) I'm gonna merge this one lickety split. I've got a goal to get to %99\\%$ coverage soon but not quite on this one.

Based on experiences from other projects, it takes about a few months and another thousand lines of test code to reach $100\\%$, which I'd like to put on the table as a reachable goal. Let's party.

mborland:
> Hi Matt (@mborland) I'm gonna merge this one lickety split. I've got a goal to get to $99\\%$ coverage soon but not quite on this one.
> 
> 
> 
> Based on experiences from other projects, it takes about a few months and another thousand lines of test code to reach $100\\%$, which I'd like to put on the table as a reachable goal. Let's party.

Sounds good to me. For the new fast types I've been recycling the same test code so hopefully it doesn't drop the coverage too much.

include/boost/decimal/detail/cmath/exp.hpp
```diff
@@ -43,20 +43,13 @@ constexpr auto exp_impl(T x) noexcept
...
-            else
-            {
-                result = x;
-            }
+            result = (signbit(x) ? zero : std::numeric_limits<T>::infinity());
         }
         else if (fpc == FP_NAN)
         {
-            result = x;
+            result = std::numeric_limits<T>::quiet_NaN();
```
mborland, line 50: I think this should stay as `result = x` so we return the correct type of NAN from the input.

include/boost/decimal/detail/cmath/exp.hpp
```diff
@@ -43,20 +43,13 @@ constexpr auto exp_impl(T x) noexcept
...
         }
-    }
+    } // LCOV_EXCL_LINE
```
mborland, line 52: Is this needed?

ckormanyos:
Well I made a few efforts to get better on funcs and also got a few lines coveredon 128/256 bit backends. I did not quite finish everything on the (non-published) TODO list. But it's headed in the right direction. So when cycles green, I'll push this thing.

Funcs are getting rather stable nowdays. Good stuff.

Issue #565, opened 2024-05-16
Add trunc/round of specified precision


mborland:
We currently offer the normal <cmath> `trunc(x)` and `round(x)`. We could additionally offer `trunc(x, num_digits)` and `round(x, num_digits)` instead of giving the user digits10 precision. These are non-standard, but could be of use in financial applications. They should also be reasonably easy to implement since we have to pull the decimal number apart into significand and exponent already. Just get rid of a few more digits and apply `detail::fenv_round(x)` at the final step

PR #586, opened 2024-05-22
Fixed width rounding

mborland:
Closes: #565 

@ckormanyos This one was by request from the discussion on the ML. I am open to a better name than overloading `trunc` even though that's what it does; truncates to a fixed number of decimal places (with appropriate rounding).


mborland:
I renamed this to trunc_to which I think more accurately describes what it does without being overly verbose

Issue #547, opened 2024-05-10
reimann_zeta


ckormanyos:

ckormanyos:
I have good, terse code for real-valued Riemann-Zeta function hanging around. I'll take a stab at this one...

Cc: @mborland

ckormanyos:
See #593

PR #593, opened 2024-05-24
Riemann zeta

ckormanyos:


ckormanyos:
See #547

ckormanyos:
I'm adjusting tolerances and probably still need a few cover lines, but in a few runs R-Zeta will be dialed

PR #624, opened 2024-05-31
elliptic integral improvements

ckormanyos:
The purpose of this PR is to get the C++17 elliptic integral functions working and tested.


ckormanyos:
- Need to do some series or perturbative method near ${\phi}=1$.
- Still need to implement and test `comp_ellint_2` and `ellint_2`.
- Still need to implement and test `comp_ellint_3` and `ellint_3` (but I don't yet have these in legacy code).
- Need $128$-bit tests all around.

Might stop at `ellint_1` and `ellint_2` complete and postpone `ellint_3`. I'll need a few days for this state anyway.

ckormanyos:
OK I got `ellint_1` and `ellint_2` and also handled with series approximation the regions ${\phi}~{\approx}~0$. Also $128$-bit tests are running.

- Still need to handle regions $k~{\approx}~1$ and edge cases.
- This PR will probably leave `ellint_3` for another time.

ckormanyos:
Hi Matt (@mborland) this one makes a pretty good start at `elliptix_x`. I'll handle either all or 2/3 of the points above and then hammer this one in within a day or so. This is a cool function to compute since it is one of the epic examples of actually using AGM-iteration to get numerical answers.

include/boost/decimal/detail/cmath/ellint_1.hpp
```diff
@@ -29,124 +30,133 @@ namespace decimal {
...
+
+  const auto fpc_m   = fpclassify(m);
+
+  if(fpc_m == FP_ZERO)
+  {
+    result = numbers::pi_v<T> / 2;
+  }
+  else if((fabs(m) > one) || (fpc_m != FP_NORMAL))
+  {
+    result = std::numeric_limits<T>::quiet_NaN();
```
mborland, line 138: Should we use qNaN or sNaN in the cmath functions in this case? I think you have been using qNaN and I have used a few sNaNs. The C++ standard just says NAN

mborland (commented):
Thanks for getting this working properly!

include/boost/decimal/detail/cmath/ellint_1.hpp
ckormanyos: Good question. I was hoping to use exclusively one or the other in the project. Of course, on bare metal signalling NaN would be difficult. So if you do intend to have any small-weight clients, I'd say use `quiet_NaN()` exclusively. But I won't make a big fuss about it either way.

ckormanyos:
> Thanks for getting this working properly!

Sure thing Matt (@mborland).

It's not _quite_ yet done. In the 32-bit case, I still have a few convergence misses (like one in a few million). I still need to fully fix the iteration parameters and see why some extreme cases don't converge. I might just use Taylor series or Pade for a sub-class of 32-bit cals.

Anyway, I need still a day or two to get this thing into develop.

include/boost/decimal/detail/cmath/ellint_1.hpp
mborland: I'll open an issue to covert to `quiet_NaN()`. Since `issignaling` isn't standard theres not even a good way to differentiate at the moment. Might as well use the one we know is good everywhere.

ckormanyos:
Still getting tiny fuzz at ${\aapprox}~\frac{1}{1,000,000}$ rate. The answer for sporadic points is a factor of $2$ too large. Re-convert PR to draft.

ckormanyos:
Hi Matt (@mborland) this one took a while. I finished `ellint_1()` and `ellint_2()` and also improved `tan()` and a few other problem-ettes along the way.

mborland (approved):
Looks good to me! Thanks for doing all this

ckormanyos:
Thanks Matt. This wqas a challenging one. I ended up using some old code and the fundamental AGM equations found in the [NIST DLMF here](https://dlmf.nist.gov/19.8). It is always fun to derive an algorithm directly from the literature.

I'll move on to `eelint_3()`, probably sticking with AGM methods.

Issue #643, opened 2024-06-10
Improve digit counting


mborland:
Rather than using the current binary search trees we could use the various bit scanning instrinsics or assembly instructions and take `floor(log10(2) * binary digits) + 1`.

PR #646, opened 2024-06-11
Improve digit counting methods

mborland:
Currently we use binary search tree to figure out the number of digits in a number. Looks like in the case of `std::uint32_t` we could covert to log calculation with a >15% speedup. I expect similar speedups for uint64_t and __uint128_t where it is more desparately needed.

Closes: #643


mborland:
Benchmarks showed improvement for the operation, but when run inside the decimal benchmarks it caused decimal32 to slow down by greater than a factor of 2. Continuing to investigate with uint64_t since that has many more operations in the search tree.

mborland:
Improvements for uint64_t were inconclusive, but using the same method for computing log on uint128_t yielded an about 8% improvement.

mborland:
Replaced by different method in: https://github.com/cppalliance/decimal/pull/647

PR #666, opened 2024-06-20
Ellint coverage

ckormanyos:


ckormanyos:
The purpose of this PR is to hit some missing `ellint_1()` and `ellint_2()` lines via adding tests and also to annotate some lines discovered that are not intended to be covered.

ckormanyos:
Better, good enough to merge but we will finish even more later

PR #711, opened 2024-07-15
decimal32 operator< without normalization in the general case

mborland:


mborland:
Provides ~20% improvement in comparison benchmarks for decimal32

PR #712, opened 2024-07-16
Test CUDA proof of concept

mborland:


mborland:
@sdarwin Same question about the CUDA runner not picking up. This PR is nearly the same as the math one but in an existing `CppAlliance` repo.

sdarwin:
Try now.

mborland:
> Try now.

Picks up now. Thanks.

sdarwin:
@mborland here's an idea.  Create a repository cppalliance/math-cuda.   It will be a mirror of boostorg/math.  Modify github actions so that in addition to jobs running from a commit or pull request, there is also a one-week schedule.   

In the CUDA gh actions, conditionally verify the current repo is cppalliance/math-cuda.    Otherwise, it just skips everything, or exits successfully.

In the CUDA gh actions, git checkout boostorg/math, and push it to the local repo which is cppalliance/math-cuda.  In other words, fast-forward merge boostorg/math -> cppalliance/math-cuda.    

That will keep the repos in sync.  It creates another commit,  causing CI to run. 

The end result is the CUDA job will run once-per-week from cppalliance/math-cuda

mborland:
> @mborland here's an idea. Create a repository cppalliance/math-cuda. It will be a mirror of boostorg/math. Modify github actions so that in addition to jobs running from a commit or pull request, there is also a one-week schedule.
> 
> In the CUDA gh actions, conditionally verify the current repo is cppalliance/math-cuda. Otherwise, it just skips everything, or exits successfully.
> 
> In the CUDA gh actions, git checkout boostorg/math, and push it to the local repo which is cppalliance/math-cuda. In other words, fast-forward merge boostorg/math -> cppalliance/math-cuda.
> 
> That will keep the repos in sync. It creates another commit, causing CI to run.
> 
> The end result is the CUDA job will run once-per-week from cppalliance/math-cuda

That makes sense. This is green now so I think I've nailed down the correct way to run NVCC with CMake testing. I need to add cmake testing support to math or also figure out with this one how to run it with B2 as well.

mborland:
NVCC is unsupported by B2 so either support needs to be merged into B2, or Cmake needs to be used

PR #723, opened 2024-08-24
Fix a typo

jk-jeon:
Found in https://github.com/jk-jeon/dragonbox/pull/66.

According to @tobybell, it appears to work even with `UINT32_C` but it's definitely a typo anyway.


mborland (approved):
Thank you for the patch!

Issue #750, opened 2024-10-04
Allow <charconv> enums and structures to be used


mborland:
Additional overloads can be added that support these when they are available. It should be just a quick alias and conversion operator to make this happen.

PR #752, opened 2024-10-04
Enable interoperability with `std::` types from `<charconv>`

mborland:
Closes: #750


mborland:
Spurious failure in test_pow. Might be worth increasing tolerance just a bit since this happens every so often.

Issue #784, opened 2025-01-08
How do I create negative numbers?


akrzemi1:
The docs aren't that clear on how to create negative numbers.

There is a declaration:

```c++
template <typename Integral1, typename Integral2>
constexpr decimal128(Integral1 coeff, Integral2 exp, bool sign = false) noexcept;
```

`sign` looks like it, but it is still not clear if `sign == true` means "positive" or "negative". Maybe name `is_negative` would be more self-explanatory.

But even then, I can also do it with `coeff`:

```c++
decimal128 d {-1, 1};
```

But then I wonder whether the following is positive or negative:

```c++
decimal128 d {-1, 1, true};
```

And then I wonder whether 

```c++
decimal128 d {-1, 1, false};
```

has the opposite meaning or not.

This constructor requires more attention in the docs.

PR #785, opened 2025-01-09
Expand the basics section

mborland:
Closes: #784 

@akrzemi1 let me know if there's anything you think is missing from the expanded basics section. I believe I hit all the points in your linked issue.

### DIFF
--- a/doc/decimal/basics.adoc
+++ b/doc/decimal/basics.adoc
@@ -1,0 +1,85 @@
+////
+Copyright 2025 Matt Borland
+Distributed under the Boost Software License, Version 1.0.
+https://www.boost.org/LICENSE_1_0.txt
+////
+
+[#basics]
+= Basic Usage
+:idprefix: basics_
+
+== Construction of Decimal Types
+
+Every decimal type can be constructed in a few ways:
+
+1) The parameters to the constructor are like so:
+
+[source, c++]
+----
+template <typename T, typename T2>
+constexpr decimal32(T coeff, T2 exp, bool sign = false) noexcept;
+----
+
+Where types `T` and `T2` are integral types (signed and unsigned are both allowed).
+Lastly the sign follows the convention of `signbit` where `false` is positive and `true` is negative.
+If both a negative coefficient and a sign are passed then the resulting decimal number will be negative.
+The final number constructed is in the form coeff x 10^exp.
+
+[souce, c++]
+----
+boost::decimal::decimal32 a {1, 0};         // constructs 1
+boost::decimal::decimal32 b {-2, 0};        // constructs -2
+boost::decimal::decimal32 c {2, 0, true};   // Also constructs -2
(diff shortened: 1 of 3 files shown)


doc/decimal/basics.adoc
```diff
@@ -0,0 +1,85 @@
...
+[source, c++]
+----
+template <typename T, typename T2>
+constexpr decimal32(T coeff, T2 exp, bool sign = false) noexcept;
+----
+
+Where types `T` and `T2` are integral types (signed and unsigned are both allowed).
+Lastly the sign follows the convention of `signbit` where `false` is positive and `true` is negative.
+If both a negative coefficient and a sign are passed then the resulting decimal number will be negative.
+The final number constructed is in the form coeff x 10^exp.
```
akrzemi1, line 26: `coeff x 10^exp` does not take into account the value of `sign`.

doc/decimal/basics.adoc
```diff
@@ -0,0 +1,85 @@
...
+#include <boost/decimal.hpp>
+#include <iostream>
+#include <iomanip>
+
+int main()
+{
+    using namespace boost::decimal;
+
+    // Outputs 0.30000000000000004
+    std::cout << std::setprecision(17) << 0.1 + 0.2;
```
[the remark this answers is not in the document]
ckormanyos: > put `std::endl` at the end

If you get exceptionally picky, you are not supposed to modify and leave modified the flags of the standard output. So yoou would actually want to output the precision-modified string to a `std::stringstream` and then print sthe stream's `str()` method and then, yes, add `std::endl;

doc/decimal/basics.adoc
```diff
@@ -23,7 +23,7 @@ constexpr decimal32(T coeff, T2 exp, bool sign = false) noexcept;
 Where types `T` and `T2` are integral types (signed and unsigned are both allowed).
 Lastly the sign follows the convention of `signbit` where `false` is positive and `true` is negative.
 If both a negative coefficient and a sign are passed then the resulting decimal number will be negative.
-The final number constructed is in the form coeff x 10^exp.
+The final number constructed is in the form (sign ? -1 : 1) x coeff x 10^exp.
```
akrzemi1, line 26: Actually, this is still incorrect, as this would render a positive decimal number when both `sign == true` and `coeff < 0`.

```c++
(sign && coeff >= 0 ? -1 : 1) x coeff x 10^exp
```

I think this also takes care for the negative zero case.

mborland: How about: 

```
(sign || coeff < 0 ? -1 : 1) x coeff x 10^exp
```

Which matches the actual constructor logic:
```
if (coeff < 0 || sign)
      {
          bits_ |= detail::d32_sign_mask;
          isneg = true;
      }
```

With (sign && coeff >= 0) the assignment (true, -1) would yield 1 when it should be -1.

PR #796, opened 2025-01-18
Fix limits and printing of subnormal_min

### LINKED ISSUE #794: Incorrect numeric_limits::max
I could be wrong, but it seems to be a bug. 

When I do:
```
--- a/include/boost/decimal/decimal64.hpp
+++ b/include/boost/decimal/decimal64.hpp
...
-    static constexpr auto (max)        () -> boost::decimal::decimal64 { return {9'999'999'999'999'999, max_exponent}; }
+    static constexpr auto (max)        () -> boost::decimal::decimal64 { return {9'999'999'999'999'999, max_exponent-16}; }
```
I get correct value:
`max: 9.99999999999999900000e+384`

with original code I get infinity.

by correct I mean what https://www.open-std.org/JTC1/SC22/WG21/docs/papers/2009/n2849.pdf specifies:

> Expansion: an rvalue of type decimal64 equal to the maximum finite number that can
> be represented by an object of type decimal64; exactly equal to 9.999999999999999 x 10^384  (there are fifteen 9's after the decimal point) 

This is just for decimal64,  others have same issue.

I think adding test that uses "excessive" precision(e.g.      `std::print("max: {:.30e}\n", nl::max());` ) to print numeric limits might be useful since I literally copy/pasted the value from PDF(
 9.999999999999999) into find of my output window and I saw that after fix above output matches(all digits after those nines were zero and exponent matched).

mborland:
Closes #794


libbooze:
I was just opening a new bug for printing... if you have time can you check if this now works as expected:

```
template<typename T>
void print_mins()
{
    std::print("{}\n", boost::typeindex::type_id<T>().pretty_name());
    using nl = std::numeric_limits<T>;
    const auto denorm_min = nl::denorm_min();
    const auto denorm_min_mult = denorm_min*T(1'000'000'000'000'000'000);
    const auto min = nl::min();
    std::print("denorm min                        : {:.30e}\n", denorm_min);
    std::print("denorm min multiplied             : {:.30e}\n", denorm_min_mult);
    std::print("min                               : {:.30e}\n\n", min);
}
```

mborland:
> I was just opening a new bug for printing... if you have time can you check if this now works as expected:
> 
> ```
> template<typename T>
> void print_mins()
> {
>     std::print("{}\n", boost::typeindex::type_id<T>().pretty_name());
>     using nl = std::numeric_limits<T>;
>     const auto denorm_min = nl::denorm_min();
>     const auto denorm_min_mult = denorm_min*T(1'000'000'000'000'000'000);
>     const auto min = nl::min();
>     std::print("denorm min                        : {:.30e}\n", denorm_min);
>     std::print("denorm min multiplied             : {:.30e}\n", denorm_min_mult);
>     std::print("min                               : {:.30e}\n\n", min);
> }
> ```

For:
```
std::cout << std::scientific
              << "denorm min: " << std::numeric_limits<decimal32>::denorm_min()
              << "\ndenorm min multiplied: " << (std::numeric_limits<decimal32>::denorm_min() * decimal32(UINT64_C(1'000'000'000'000'000'000)))
              << "\nmin: " << std::numeric_limits<decimal32>::min() << std::endl;
```

we get:
```
denorm min: 1.000000e-101
denorm min multiplied: 1.000000e-83
min: 1.000000e-95
```

Denorms were being treated as non-finite values.

libbooze:
> we get:
> 
> ```
> denorm min: 1.000000e-101
> denorm min multiplied: 1.000000e-83
> min: 1.000000e-95
> ```
> 
> Denorms were being treated as non-finite values.

Nice, thank you for the fix.

test/test_limits.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+#include <boost/decimal.hpp>
+#include <boost/core/lightweight_test.hpp>
+
+#if defined(__cpp_consteval) && __cpp_consteval >= 201811L && !defined(BOOST_DECIMAL_NO_CONSTEVAL_DETECTION)
+
+using namespace boost::decimal;
+
+template <typename T>
+consteval bool test_value(T value, const char* res)
```
libbooze, line 13: Would passing res as string_view here made logic a bit simpler? 
Then you would only need to make the string_view from (buff, r.ptr) and return if they are same?

test/test_limits.cpp
```diff
@@ -0,0 +1,122 @@
...
+        if (*buffer_ptr++ != *res_ptr++)
+        {
+            return false;
+        }
+    }
+    
+    return true;
+}
+
+int main()
```
libbooze, line 32: maybe testing that `lowest + max == 0` is worthwhile test to add?

test/test_limits.cpp
```diff
@@ -0,0 +1,122 @@
...
+    static_assert(test_value(std::numeric_limits<decimal64_fast>::epsilon(), "1.0000000000000000000000000000000000000000e-15"));
+    static_assert(test_value(std::numeric_limits<decimal128>::epsilon(), "1.0000000000000000000000000000000000000000e-33"));
+    static_assert(test_value(std::numeric_limits<decimal128_fast>::epsilon(), "1.0000000000000000000000000000000000000000e-33"));
+
+    // Min
+    static_assert(test_value(std::numeric_limits<decimal32>::min(), "1.0000000000000000000000000000000000000000e-95"));
+    static_assert(test_value(std::numeric_limits<decimal32_fast>::min(), "1.0000000000000000000000000000000000000000e-95"));
+    static_assert(test_value(std::numeric_limits<decimal64>::min(), "1.0000000000000000000000000000000000000000e-383"));
+    static_assert(test_value(std::numeric_limits<decimal64_fast>::min(), "1.0000000000000000000000000000000000000000e-383"));
+    static_assert(test_value(std::numeric_limits<decimal128>::epsilon(), "1.0000000000000000000000000000000000000000e-33"));
```
libbooze, line 55: is this c/p mistake?

mborland:
@libbooze I believe I hit all of your comments

test/test_limits.cpp
```diff
@@ -64,6 +56,13 @@ int main()
     static_assert(test_value(std::numeric_limits<decimal128>::denorm_min(), "1.0000000000000000000000000000000000000000e-6176"));
     static_assert(test_value(std::numeric_limits<decimal128_fast>::denorm_min(), "1.0000000000000000000000000000000000000000e-6143"));
 
+    static_assert(test_value(std::numeric_limits<decimal32>::max() + std::numeric_limits<decimal32>::lowest(), "0.0000000000000000000000000000000000000000e+00"));
```
libbooze, line 59: nit:  add same comment(// Lowest + max) as in other preprocessor branch to keep it consistent

libbooze:
> @libbooze I believe I hit all of your comments

yes, thank you. Left just one comment regarding comment.

mborland:
> > @libbooze I believe I hit all of your comments
> 
> 
> 
> yes, thank you. Left just one comment regarding comment. 
> 
> 

I'll add it. This needs a little more investigation because I think something has gone wrong with fpclassify now. The fuzzer seems like it's consistently hitting an assertion.

libbooze:
Not sure if this is related to this issue, so feel free to not address this in this pull, but I noticed another issue with printing. Just mentioning it in case you might find it useful.
(this is with latest commit 53b0c3d81916a970a23fca9d6c08407fbafb80bf  from this branch):


```
    std::print("{:.5e}\n", std::numeric_limits<decimal32>::max());
    std::print("{:.6e}\n\n", std::numeric_limits<decimal32>::max());
    std::print("{:.14e}\n", std::numeric_limits<decimal64>::max());
    std::print("{:.15e}\n\n", std::numeric_limits<decimal64>::max());
    std::print("{:.32e}\n", std::numeric_limits<decimal128>::max());
    std::print("{:.33e}\n\n", std::numeric_limits<decimal128>::max());
```

on my machine prints:

```
9.9999990000000000000000000000000000000000e+96
9.9999999999999990000000000000000000000000e+384
9.9999999999999999999999999999999990000000e+6144

1.000000e+96
9.999999e+96

1.000000000000000e+384
9.999999999999999e+384

1.000000000000000000000000000000000e+6144
9.999999999999999999999999999999999e+6144
```

mborland:
Can you open as a separate issue? I think this one has already exceeded its original scope since it wipes out a few issues. I'll have to compare <print> to regular io streams to see if they're identically wrong. The <format> support is the newest part of the library and got merged in like a week before review.

libbooze:
> Can you open as a separate issue? I think this one has already exceeded its original scope since it wipes out a few issues. I'll have to compare print to regular io streams to see if they're identically wrong. The format support is the newest part of the library and got merged in like a week before review.

Sure, but will wait till this is merged so I can test on develop. Easier to track/reproduce then.

Issue #802, opened 2025-01-19
Incorrect result range check in d128_mul_impl


libbooze:
In `d128_mul_impl` there is check to see if we can compute product without overflowing, but it is incorrect. 

If we make this change to source:
```
     else
     {
+        std::cerr << "can not mult in 128bit because " <<  (lhs_dig * rhs_dig) << " > "  << std::numeric_limits<uint128>::digits10 << std::endl;
         // Once we have the normalized significands and exponents all we have to do is
         // multiply the significands and add the exponents

```
And then run this code:

```
void fasting()
{
    using namespace boost::decimal;
    using Dec= decimal128;
    std::print("{}\n", boost::typeindex::type_id<Dec>().pretty_name());
    const Dec a{100'000,1};  // 6 dec digits significant
    const Dec b{2'000'000,1}; // 7 dec digits significant
    std::print("a  : {:.20e}\n", a);
    std::print("b  : {:.20e}\n", b);
    std::print("a*b: {:.20e}\n", a *b );
}
```
we get cerr output
`can not mult in 128bit because 42 > 38`

I guess obvious fix is to just replace multiplication with addition.

Beside that after this is fixed I wonder why is code calling `num_digits` just to compare sum(currently product) with digits10 of binary number. I would presume getting binary digits of both numbers and comparing sum to 128 is faster.

PR #812, opened 2025-01-21
Fix calculation of resultant digits to keep us in the range of `uint128`

mborland:
Closes: #802

### DIFF
--- a/test/github_issue_802.cpp
+++ b/test/github_issue_802.cpp
@@ -1,0 +1,28 @@
+// Copyright 2025 Matt Borland
+// Distributed under the Boost Software License, Version 1.0.
+// https://www.boost.org/LICENSE_1_0.txt
+
+#include <boost/decimal.hpp>
+#include <boost/core/lightweight_test.hpp>
+#include <iostream>
+
+void fasting()
+{
+    using namespace boost::decimal;
+    using Dec = decimal128;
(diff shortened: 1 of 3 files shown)


test/github_issue_802.cpp
```diff
@@ -0,0 +1,28 @@
+// Copyright 2025 Matt Borland
+// Distributed under the Boost Software License, Version 1.0.
+// https://www.boost.org/LICENSE_1_0.txt
+
+#include <boost/decimal.hpp>
+#include <boost/core/lightweight_test.hpp>
+#include <iostream>
+
+void fasting()
```
libbooze, line 9: Not sure adding this as test is useful. 
Result was always correct afaik.
This is hard to test but if you want one way to test this is to introduce detail:: helper that checks if 2 numbers can be multiplied within range of u128, and then test that helper. 
Might be an overkill, but it is basically just extracting
```
    const auto lhs_dig {detail::num_digits(lhs_sig)};
    const auto rhs_dig {detail::num_digits(rhs_sig)};
    return lhs_dig + rhs_dig <= std::numeric_limits<uint128>::digits10;
    
```


into bool returning helper function.

mborland: It's a good reproducer for using the debugger. I'm not super worried about the test in this case since the special functions test suites will for sure hit both paths.

Issue #832, opened 2025-01-29
Boost Review: Are concept names public?


jzmaddock:
Are concept names supposed to be public? If they are, please
document them as regular entities. If they are not, please move them
into the detail namespace.

PR #847, opened 2025-01-31
Move concepts into detail namespace

mborland:
Closes: #832


mborland:
All failures are from a -Wundef in Boost.Random which should now be fixed.

Issue #836, opened 2025-01-29
Boost Review: undocumented public constructor


jzmaddock:
There is an undocumented public constructor: constexpr decimal32(bool coeff, T exp, bool sign = false) noexcept; which appears to be a mistaken addition?

PR #849, opened 2025-01-31
Remove decimal32 bool constructor

mborland:
Closes: #836


mborland:
Failures are from Boost.Random -Wundef

PR #857, opened 2025-02-13
Better Addition

mborland:
This provides a 1.5 - 2x speedup for addition in addition to reducing code duplication.


mborland:
I think there's something ever so slightly wrong in here but I can't pick it out. Theoretically we should be able to do lhs + (-rhs) to implement subtraction, but with this changeset sin, cosine, and zeta tests all begin to fail. Error goes from under 35 ULPS to order 10,000 ULPs

Issue #844, opened 2025-01-29
Boost Review: code duplication part 2.


jzmaddock:
Code duplication: Importantly, I don't understand why the trait `bias_v` is even needed, given that it duplicates the value stored in `std::numeric_limits<...>::digits`/`digits10`/`maxdigits10`. Same story with the `emin_v` and `emax_v` traits compared to the `std::numeric_limits<...>::min_exponent`/`min_exponent10` and `std::numeric_limits<...>::max_exponent`/`max_exponent10`. This seems like some code duplication that would be prone to causing confusion later. Why do multiple definitions of traits exist that have the same values? There may be other cases of this as well. I would love to see either `std::numeric_limits` versions being defined in terms of the other traits where possible, or vice-versa. I find it confusing to have 2 bespoke definitions.

mborland:
I think this will be closed as well by https://github.com/cppalliance/decimal/pull/863. We can't replace the traits with std::numeric_limits since the traits are used in the implementation of the classes. The traits are used to define our values in std::numeric_limits so there is consistency there.

Issue #842, opened 2025-01-29
Boost Review: Rationalize traits class usage to make the code easier to read.


jzmaddock:
(1) Sometimes there are traits used, such as `detail::bias_v<>`, but other times the trait's functionality is replicated in-line. I find this more difficult to read, instead of using the trait. As an example, in the function `detail::to_chars_impl()`, we have this line:
```
constexpr auto max_fractional_value = std::is_same<TargetDecimalType, decimal32>::value || std::is_same<TargetDecimalType, decimal32_fast>::value ? TargetDecimalType{1, 7} : std::is_same<TargetDecimalType, decimal64>::value || std::is_same<TargetDecimalType, decimal64_fast>::value ? TargetDecimalType{1, 16} : TargetDecimalType{1, 34};
```
This line could easily be replaced with the following, which means exactly the same thing, but is easier to visually parse.
```
constexpr auto max_fractional_value = TargetDecimalType{1, bias_v<TargetDecimalType>};
```
I don't want this to sound like a nit-pick, but cases like this have made it more difficult for me to understand and work through the code.

(2) On the topic of traits, I think there should be some more traits defined, to remove usage of things like `std::is_same<...>::value`. This is another thing that has made it more difficult for me to read the code. Something along the lines of `detail::is_fast<...>` to denote whether you have a `_fast` type or not. For example, you could replace the following
```
std::enable_if_t<std::is_same<DecimalType, decimal32>::value || std::is_same<DecimalType, decimal64>::value || std::is_same<DecimalType, decimal128>::value, bool>
```
With the following
```
std::enable_if_t<!detail::is_fast<DecimalType>::value, bool>
```
As well, `detail::bit_width<...>` would help remove lines like `std::is_same<T, decimal128>::value || std::is_same<T, decimal128_fast>::value`, and turn them into `detail::bit_width<...>::value == 128`. These are just suggestions and don't need to be taken verbatim. Anything to remove duplication and make the code easier to understand.

PR #863, opened 2025-02-17
Use traits to reduce duplication / improve logic 

mborland:
Closes: #842 
Closes: #844


mborland:
So clang and MSVC have this figured out by GCC prior to 12 has issues with incomplete types...

PR #884, opened 2025-02-27
128-Bit division simplifications

mborland:


mborland:
Windows failures look like a passing bug in the CI system since it now says chrono not found, but it runs fine in drone.

Issue #885, opened 2025-02-28
Rework u128


mborland:
Right now it has macros to define operators and is entirely missing a bunch of comparison operators. This can also be the time to explore inserting intrinsics into say the mul operator

mborland:
Some results on ARM64 where old is our current emulated128 and new is the one I'm building now:

```
---------------------------
Two Word Operations
---------------------------

comparisons<builtin    >: 141153     us (s=299999985)
comparisons<old        >: 166175     us (s=299999985)
comparisons<new        >: 128813     us (s=299999985)

add<Builtin    >: 20494      us (s=6185515908288643546)
add<Old        >: 19391      us (s=6185515908288643546)
add<New        >: 18791      us (s=6185515908288643546)

sub<Builtin    >: 20559      us (s=6843120887274356308)
sub<Old        >: 19733      us (s=6843120887274356308)
sub<New        >: 18890      us (s=6843120887274356308)

mul<Builtin    >: 20399      us (s=4541227745721732167)
mul<Old        >: 19099      us (s=4541227745721732167)
mul<New        >: 18995      us (s=4541227745721732167)

---------------------------
One Word Operations
---------------------------

comparisons<builtin    >: 141399     us (s=299999985)
comparisons<old        >: 162425     us (s=299999985)
comparisons<new        >: 126388     us (s=299999985)

add<Builtin    >: 24094      us (s=7061247353260042742)
add<Old        >: 19478      us (s=7061247353260042742)
add<New        >: 18831      us (s=7061247353260042742)

sub<Builtin    >: 20488      us (s=18357599163646591540)
sub<Old        >: 19613      us (s=18357599163646591540)
sub<New        >: 18680      us (s=18357599163646591540)

mul<Builtin    >: 20308      us (s=1771533121646374909)
mul<Old        >: 19077      us (s=1771533121646374909)
mul<New        >: 19075      us (s=1771533121646374909)
```

mborland:
Here are x64 results:

```
---------------------------
Two Word Operations
---------------------------

comparisons<builtin    >: 153029     us (s=299999985)
comparisons<old        >: 144080     us (s=299999985)
comparisons<new        >: 73615      us (s=299999985)

add<Builtin    >: 33656      us (s=6185515908288643546)
add<Old        >: 32987      us (s=6185515908288643546)
add<New        >: 32876      us (s=6185515908288643546)

sub<Builtin    >: 33735      us (s=6843120887274356308)
[...]

PR #887, opened 2025-03-13
U128 v2

mborland:
Closes: #885


mborland:
@ckormanyos since this PR is already a monster I think I'm going to merge it once it turns green, and then do the integration work into a separate PR. There is also some demand for a boost library containing just a u128 and i128. I think it would be pretty easy to copy-paste this into such a thing.

ckormanyos:
This is, indeed, a lot of work and I need to catch up.

> There is also some demand for a boost library containing just a u128 and i128. I think it would be pretty easy to copy-paste this into such a thing.

I believe so. We have considered this in some form or another in numerous offline chats. I guess (u)int128 would be something like a prerequisite for (not only this, but also other) library(ies).

In the old days, we did some work with `float128` in both math and multiprecision without going the formal route for a full-on library. It might be an option to make (u)int128 in an optimized form be something within a larger context of an existing header-only library?

mborland:
> This is, indeed, a lot of work and I need to catch up.
> 
> > There is also some demand for a boost library containing just a u128 and i128. I think it would be pretty easy to copy-paste this into such a thing.
> 
> I believe so. We have considered this in some form or another in numerous offline chats. I guess (u)int128 would be something like a prerequisite for (not only this, but also other) library(ies).
> 
> In the old days, we did some work with `float128` in both math and multiprecision without going the formal route for a full-on library. It might be an option to make (u)int128 in an optimized form be something within a larger context of an existing header-only library?

I had previously talked with Peter Dimov about adding uint128 and int128 to core, but since this PR and an equivalent int128 would take over 11,000 lines he thinks it's best separate. A few others have said the same since these additions would be an almost 50% increase in the size of core.

mborland:
@ckormanyos do you want me to wait till you have a chance to look at this before merging? It's not really stopping anything since I am now looking into SIMD for u256.

ckormanyos:
> @ckormanyos do you want me to wait till you have a chance to look at this before merging?

Thank you Matt (@mborland). This was a very large effort on your part. I've followed it at least insofar as I could. I think it is cool and I like the Knuth long-div adaption.

Whenever you are OK with it, I'm fine with it.

mborland:
This work is being ported into the repo: https://github.com/cppalliance/int128 for broader use.

Issue #888, opened 2025-03-13
Rework u256


mborland:
This can be less comprehensive since we only need a few operations from the type, but can use the ADX instructions and others to speed up a few of the operations?

mborland:
AVX2 seems like the way to go for some instructions since we can jam the std::uint64_t vals[4]; into `__m256i` and perform operations that way

mborland:
Here's some early results on macOS ARM64:

```
---------------------------
Four Word Operations
---------------------------

comp<old        >: 495481     us (s=999999995)
comp<new        >: 364551     us (s=999999995)

add <Old        >: 352063     us (s=13684479268969042895)
add <New        >: 357595     us (s=13684479268969042895)

mul <Old        >: 6253920    us (s=17124140290376989252)
mul <New        >: 21192046   us (s=17124140290376989252)

div <Old        >: 20180534   us (s=9087210595)
div <New        >: 15066392   us (s=9087210595)

digits<old        >: 5839005    us (s=77040462080)
digits<new        >: 5266650    us (s=57775353110)
```

The div speedup is big since that is the preponderance of the operations we do with u256. The mul numbers are really bad in comparison though. I assume everything is 4x4 words which is probably killing us there

ckormanyos:
Hi Matt (@mborland) we need to look into this. Multiplication is the $80\\%$ operation in a typical software user-defined numeric type.

Now I forgot if we did or did not unroll the loops in 8-by-8 multiplication.

There are also special cases to handle such as four-by-four goses to 8 limbs. I might not have time to fully digest the depth of your first results, but wee might need to look into this.

Usually speeding up division at the cost of slowing down multiplication is a negative gain for overall performance. But each case is different.

ckormanyos:
>  This would allow us to drop wide-integer

I definitely want to drop the pulled-out wide-integer algorithms. These were a bit of a transitional step in the first place.

But that is easier said than done, I get that.

mborland:
I will absolutely keep poking at it. All I did was use the routines out of int128 and set a template size of 8 by 8 for both mul and div and give it a first shot. Happy with div. Not so happy with mul.

ckormanyos:
So it is a bit of a toss-up what will be overall the fastests for $256$.

Yes keep plugging away at it. I never fisnished that.

mborland:
[...]

PR #899, opened 2025-06-27
U256 v3

mborland:
Closes: #888


mborland:
Note I haven't done performance testing yet. Everything is correct locally which is a good sign. Once we know for certain the mul and div are more performant than what we already have we can do integration (likely in a different PR)

PR #937, opened 2025-07-28
Handle unclear code from local metal benching

ckormanyos:
Hi Matt (@mborland) this PR handles some warnings that I believe pointed out some unclear code. I detected these when doing local bare-metal benchmarking of the newer, faster `decimal64_t`. I ran into some (I believe) legitimate warnings on the metal and handled them.

The results of the benchmark are excellent and `decimal64_t` is better than two-fold faster than last year.

The table number was $490{\mu}s$ last year and is now $200{\mu}s$.

These are the updated numbers from boost::decimal Git-hash 0b474bf05f339f21d87dd1902dc6294bb9914254.

| Type                           |  runtime ${\mu}s$ |  relative  |    code-size [kb] |
|--------------------------------|---------------|------------|-------------------|
| 64-bit `double` (built-in, no FPU)      |  $22$          |   $1.0$      |       $5.6$         |
| ::math::softfloat::float64_t   |   $27$          |   $1.2$      |       $8.5$         |
| boost::decimal::decimal64_t    |  $200$          |   $9.1$      |       $20$          |

```
#if 0
GCC14.2, arm-none-eabi, float-abi=soft, -O2

Calculation exp() via Pade approximation approx. 15 decimal digits.
Code-size includes about 2kb for startup, clock-init, skinny MCAL and cooperative time-scheduler.

target-specific flags
---------------------
-O2
-mcpu=cortex-m4
-mtune=cortex-m4
-mthumb
-mfloat-abi=soft
-finline-functions
-finline-limit=128
-mno-unaligned-access
-mno-long-calls
#endif
```

### DIFF
--- a/include/boost/decimal/decimal32_fast.hpp
+++ b/include/boost/decimal/decimal32_fast.hpp
@@ -1000,12 +1000,16 @@ constexpr auto div_impl(decimal_fast32_t lhs, decimal_fast32_t rhs, decimal_fast
               << "\nexp rhs: " << exp_rhs << std::endl;
     #endif
 
+    using local_signed_exponent_type = std::common_type_t<std::int_fast32_t, int>;
+
+    static_assert(sizeof(local_signed_exponent_type) >= 4, "Error in local exponent type definition");
+
     // We promote to uint64 since the significands are currently 32-bits
     // By appending enough zeros to the LHS we end up finding what we need anyway
     constexpr auto ten_pow_precision {detail::pow10(static_cast<std::uint_fast64_t>(detail::precision_v<decimal32_t>))};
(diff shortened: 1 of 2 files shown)


mborland:
> Matt should I rerun the failed test on clang Darwin? Compilation itself seg-faulted on a `log1p` test file. Is this error known, bogus, sporadic, real?

Can't say it's known to me. There are sporadic failures of the special functions which for results slightly out of tolerance. I normally see it with pow.

ckormanyos:
> Can't say it's known to me. 

I started a rerun of that single job only on GHA.

mborland:
> > Can't say it's known to me.
> 
> I started a rerun of that single job only on GHA.

No issues locally on Mac so hopefully a fluke

PR #993, opened 2025-08-15
Try using Boost.CI provided coverage tool for easier maintenance

mborland:


mborland:
This does not seem better since MSVC does not obey the // LCOV_EXCL_START/STOP

PR #1030, opened 2025-09-04
Implement check for sticky bit to fix rounding for all types and implement for `decimal32_t`

mborland:
Closes: #1027

### DIFF
--- a/include/boost/decimal/decimal32_t.hpp
+++ b/include/boost/decimal/decimal32_t.hpp
@@ -617,29 +617,7 @@ constexpr decimal32_t::decimal32_t(T1 coeff, T2 exp, bool sign) noexcept // NOLI
     auto biased_exp {static_cast<int>(exp + detail::bias)};
     if (coeff > detail::d32_max_significand_value || biased_exp < 0)
     {
-        coeff_digits = detail::num_digits(coeff);
-
-        // How many digits need to be shifted?
-        const auto shift_for_small_exp {(-biased_exp) - 1};
-        const auto shift_for_large_coeff {(coeff_digits - detail::precision) - 1};
-        const auto shift {std::max(shift_for_small_exp, shift_for_large_coeff)};
-
-        // Do shifting
-        const auto shift_pow_ten {detail::pow10(static_cast<T1>(shift))};
-        const auto shifted_coeff {coeff / shift_pow_ten};
-        const auto trailing_digits {coeff % shift_pow_ten};
-
-        coeff = shifted_coeff;
-        const auto sticky {trailing_digits != 0u};
-        exp += shift;
-        biased_exp += shift;
-        coeff_digits -= shift;
-
-        // Do rounding
-        auto removed_digits = detail::fenv_round(coeff, sign, sticky);
-        exp += removed_digits;
-        biased_exp += removed_digits;
-        coeff_digits -= removed_digits;
+        coeff_digits = detail::coefficient_rounding<decimal32_t>(coeff, exp, biased_exp, sign);
     }
 
     auto reduced_coeff {static_cast<significand_type>(coeff)};
--- a/include/boost/decimal/decimal32_t.hpp
+++ b/include/boost/decimal/decimal32_t.hpp
@@ -604,22 +604,10 @@ template <BOOST_DECIMAL_UNSIGNED_INTEGRAL T1, BOOST_DECIMAL_INTEGRAL T2>
 #else
 template <typename T1, typename T2, std::enable_if_t<detail::is_unsigned_v<T1> && detail::is_integral_v<T2>, bool>>
 #endif
-constexpr decimal32_t::decimal32_t(T1 coeff_init, T2 exp, bool sign) noexcept // NOLINT(readability-function-cognitive-complexity,misc-no-recursion)
+constexpr decimal32_t::decimal32_t(T1 coeff, T2 exp, bool sign) noexcept // NOLINT(readability-function-cognitive-complexity,misc-no-recursion)
 {
     static_assert(detail::is_integral_v<T1>, "Coefficient must be an integer");
     static_assert(detail::is_integral_v<T2>, "Exponent must be an integer");
-
-    using unsigned_coeff_type = detail::make_unsigned_t<T1>;
-
-    unsigned_coeff_type coeff {};
-    BOOST_DECIMAL_IF_CONSTEXPR (detail::is_signed_v<T1>)
-    {
-        coeff = static_cast<unsigned_coeff_type>(coeff_init < 0 ? -coeff_init : coeff_init);
-    }
-    else
-    {
-        coeff = static_cast<unsigned_coeff_type>(coeff_init);
-    }
 
     bits_ = sign ? detail::d32_sign_mask : UINT32_C(0);
 
@@ -691,8 +679,8 @@ constexpr decimal32_t::decimal32_t(T1 coeff_init, T2 exp, bool sign) noexcept //
     }
 
     // If the exponent fits we do not need to use the combination field
-    const auto biased_exp {static_cast<std::uint32_t>(exp + detail::bias)};
-    if (biased_exp <= detail::d32_max_biased_exponent)
+    const auto biased_exp {static_cast<std::int32_t>(exp + detail::bias)};
+    if (BOOST_DECIMAL_LIKELY(biased_exp > 0) && biased_exp <= detail::d32_max_biased_exponent)
     {
         if (big_combination)
         {
--- a/include/boost/decimal/decimal32_t.hpp
+++ b/include/boost/decimal/decimal32_t.hpp
@@ -613,8 +613,7 @@ constexpr decimal32_t::decimal32_t(T1 coeff, T2 exp, bool sign) noexcept // NOLI
     {
         bool sticky {detail::find_sticky_bit(coeff, exp, detail::bias_v<decimal32_t>)};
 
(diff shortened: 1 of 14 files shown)


morinmorin:
That was quick!

Fails with: 
```c++
cout << boost::decimal::decimal32_t(8400000U, -102) << endl; // => 8.4e-96 (actual: 0)
```
Maybe off-by-one error?

morinmorin:
I ran a subset of Intel's testcases. Dozens of them are failing, with expected results being small non-zero numbers.
For example,
```
add(-0, 8.4e-100) =
(expect) 8.4e-100
(actual) 8e-100

mul(1, 1e-101) =
(expect) 1e-101
(actual) 0

div(1e-101, 1) =
(expect) 1e-101
(actual) 0

div(5.24289e-96, 1) =
(expect) 5.24289e-96
(actual) 5.2429e-96
```

morinmorin:
Diff for `decimal32_t.hpp`
```diff
@@ -626,11 +626,10 @@ constexpr decimal32_t::decimal32_t(T1 coeff_init, T2 exp, bool sign) noexcept //
     // If the coeff is not in range, make it so
     // Only count the number of digits if we absolutely have to
     int coeff_digits {-1};
-    if (coeff > detail::d32_max_significand_value)
+    if (coeff > detail::d32_max_significand_value || (exp + detail::bias < 0))
     {
         bool sticky_bit {detail::find_sticky_bit(coeff, exp, detail::bias_v<decimal32_t>)};
 
-        if (!sticky_bit)
         {
             coeff_digits = detail::num_digits(coeff);
             if (coeff_digits > detail::precision + 1)
@@ -660,12 +659,6 @@ constexpr decimal32_t::decimal32_t(T1 coeff_init, T2 exp, bool sign) noexcept //
                 exp += static_cast<T2>(detail::fenv_round(coeff, sign, sticky_bit));
             }
         }
-        else
-        {
-            // This should already be handled in find_sticky_bit
-            BOOST_DECIMAL_ASSERT((coeff >= 1'000'000U && coeff <= 9'999'999U) || coeff == 0U);
-            exp += static_cast<T2>(detail::fenv_round(coeff, sign, sticky_bit));
-        }
     }
 
     auto reduced_coeff {static_cast<significand_type>(coeff)};
```

Diff for `fenv_rounding.hpp`
```diff
@@ -147,24 +147,17 @@ BOOST_DECIMAL_FORCE_INLINE constexpr auto find_sticky_bit(T1& coeff, T2& exp, co
 
     if (biased_exp < 0)
     {
-        const auto shift {static_cast<unsigned>(-biased_exp)};
+        const auto shift {static_cast<unsigned>(-biased_exp) - 1};
         const auto shift_p10 {detail::pow10<T1>(shift)};
-        const auto shift_p10_min_1 {shift_p10 / 10U};
 
         // TODO(mborland): in the uint128_t case we should expose a div_mod since the mod is already available
         const auto q {coeff / shift_p10};
         const auto rem {coeff % shift_p10};
 
-        auto guard_digits {rem / shift_p10_min_1};
-        sticky = (rem % shift_p10) != 0U;
+        sticky = (rem != 0U);
 
         coeff = q;
         exp += static_cast<int>(shift);
-
-        if (guard_digits > 5U || (guard_digits == 5U && (sticky || (static_cast<std::uint64_t>(coeff) & 1U))))
-        {
-            ++coeff;
-        }
     }
 
     return sticky;
```

mborland:
Since the maximum significand being divided is 10^14 which is less than 2^51 this could be done with doubles and then casted down to integer with proper rounding.

mborland:
> Thanks, confirmed that the fix (for decimal32) works fine.
> 
> > Since the maximum significand being divided is 10^14 which is less than 2^51 this could be done with doubles and then casted down to integer with proper rounding.
> 
> Right. Or, in this specific testcase, changing `detail::pow10(static_cast<std::uint64_t>(detail::precision))` (in `div_impl.hpp`) to `detail::pow10(static_cast<std::uint64_t>(detail::precision)+1u)};` would resolve the failure.
> 
> About the `if (!BOOST_DECIMAL_IS_CONSTANT_EVALUATED(coeff))` block in `find_sticky_bit`: `find_sticky_bit` is responsible for computing `sticky` AND shifting `coeff`. So even if `sticky` doesn't need to be computed, shifting still needs to be done. There are two approaches to fix it.
> 
> * remove this block entirely; or
> * add code for updating `coeff` and `exp` inside the inner `if` block (i.e. `if (BOOST_DECIMAL_UNLIKELY(…))`.

This is already handled in the constructor: https://github.com/cppalliance/decimal/blob/develop/include/boost/decimal/decimal32_t.hpp#L616. If there's no need to find the sticky bit we can divide and round like normal. The method of finding the sticky bit is slower than just dividing by a power of 10 looked up from a table. We could even remove the modulo added in this PR in that case.

morinmorin:
Without removing `if (!BOOST_DECIMAL_IS_CONSTANT_EVALUATED(coeff))` block, this fails:
```
fesetround(rounding_mode::fe_dec_to_nearest_from_zero);
cout << decimal32_t(100000, -105) << endl; // => 1e-100 (actual: 0)
```

But anyway, this is a simpler implementation of shifting and rounding (which doesn't need `find_sticky_bit`):
```c++
int coeff_digits {-1};
auto biased_exp {static_cast<int>(exp + detail::bias)};
if (coeff > detail::d32_max_significand_value || biased_exp < 0) {
    coeff_digits = detail::num_digits(coeff);

    // How many digits need to be shifted?
    auto shift_for_small_exp = (-biased_exp) - 1;
    auto shift_for_large_coeff = (coeff_digits - detail::precision) - 1;
    auto shift = std::max(shift_for_small_exp, shift_for_large_coeff);

    // Do shifting
    auto [shifted_coeff, trailing_digits] = div_mod_pow10(coeff, shift);
    coeff = shifted_coeff;
    auto sticky {trailing_digits != 0u};
    exp += shift;
    biased_exp += shift;
    coeff_digits -= shift;

    // Do rounding
    auto removed_digits = detail::fenv_round(coeff, sign, sticky);
    exp += removed_digits;
    biased_exp += removed_digits;
    coeff_digits -= removed_digits;
}
```
In this code, two shifts are merged into one. To further improve efficiency, the code uses **division by constant** (which is done by `div_mod_pow10(x, k)`), since that's optimized well by compilers. Function `div_mod_pow10(x, k)` computes `x / (10**k)` as `x / CompileTimeConstant`. Here is a draft code (this is a manual coding, but it can be avoided using meta-programming):
```c++
template<typename Uint, unsigned K>
consteval Uint pow10()
{
    Uint x = 1;
    for (unsigned i = 0; i < K; ++i) x *= 10;
    return x;
}
constexpr std::pair<std::uint32_t, std::uint32_t> div_mod_pow10(std::uint32_t x, unsigned k)
{
    using T = std::uint32_t;
    switch(k) {
        case (0): return {x / pow10<T, 0u>(), x % pow10<T, 0u>()};
        case (1): return {x / pow10<T, 1u>(), x % pow10<T, 1u>()};
        case (2): return {x / pow10<T, 2u>(), x % pow10<T, 2u>()};
        case (3): return {x / pow10<T, 3u>(), x % pow10<T, 3u>()};
        case (4): return {x / pow10<T, 4u>(), x % pow10<T, 4u>()};
        case (5): return {x / pow10<T, 5u>(), x % pow10<T, 5u>()};
        case (6): return {x / pow10<T, 6u>(), x % pow10<T, 6u>()};
        case (7): return {x / pow10<T, 7u>(), x % pow10<T, 7u>()};
        case (8): return {x / pow10<T, 8u>(), x % pow10<T, 8u>()};
        case (9): return {x / pow10<T, 9u>(), x % pow10<T, 9u>()};
        default: return {0, x};
    }
}
```
Another option is to provide only `div_pow10` and to compute the remainder as `x - quotient * (10**k)`.

morinmorin:
Thanks, the failures in the subset of Intel's testcases are now resolved!

After adding more Intel's testcases (specifically the tests with non-`to_nearest` rounding modes), I found some failures. The failing testcase is essentially this:
```c++
boost::decimal::fesetround(boost::decimal::rounding_mode::fe_dec_downward);
cout << "5e+50"_DF - "4e+40"_DF << endl;
// (expect) 4.999999e+50
// (actual) 5e+50

boost::decimal::fesetround(boost::decimal::rounding_mode::fe_dec_upward);
cout << "5e+50"_DF + "4e+40"_DF << endl;
// (expect) 5.000001e+50
// (actual) 5e+50
```

I'm wondering if a "sticky bit"-like flag is needed in addition and subtraction.

mborland:
> Thanks, the failures in the subset of Intel's testcases are now resolved!
> 
> After adding more Intel's testcases (specifically the tests with non-`to_nearest` rounding modes), I found some failures. The failing testcase is essentially this:
> 
> ```c++
> boost::decimal::fesetround(boost::decimal::rounding_mode::fe_dec_downward);
> cout << "5e+50"_DF - "4e+40"_DF << endl;
> // (expect) 4.999999e+50
> // (actual) 5e+50
> 
> boost::decimal::fesetround(boost::decimal::rounding_mode::fe_dec_upward);
> cout << "5e+50"_DF + "4e+40"_DF << endl;
> // (expect) 5.000001e+50
> // (actual) 5e+50
> ```
> 
> I'm wondering if a "sticky bit"-like flag is needed in addition and subtraction.

I believe these are in the same boat as division where additional offset can easily be applied internally. Division was extending the lhs argument to 14 integer digits, but since the type is `std::uint64_t` it really costs nothing to extend it to `std::numeric_limits<std::uint64_t>::digits10`. Same kind of shift based alignment is occurring in both add and sub.

PR #1047, opened 2025-09-10
Add doc search

mborland:


sdarwin:
I might not have seen the most recent comment:  

"I think this is what needs to be added to boostlook: https://gitlab.com/antora/antora-lunr-extension#user-content-search-input"  

Yes, it may be that should be added to boostlook, and/or this target library (decimal).  

To include an extension, place it in package.json, package-lock.json, and the playbook.yml file.

An issue is that boostlook would apply to all boostlook users. Right?  Do they all require antora-lunr-extension?  The website already offers search.   From this logic, maybe it should just be added in decimal.

PR #1070, opened 2025-09-19
Try re-adding older compiler support

mborland:


mborland:
One version down of Clang and GCC are both hopelessly broken

PR #1076, opened 2025-09-24
Add CI testing of modules

mborland:
Locally, GCC-15 allows us to build and link to the module using regular b2 requirements descriptions. All tests should be trivially changed to support testing with either the normal include, or importing the module.

@anarthal I don't know if this is of interest to you or not, but GCC-15 is the first time I've been able to trivially run all the tests like you normally would from the test directory. `import std` gives me an ICE so I include all the normal headers before the module.


anarthal:
Yes, this is interesting, specially the way you're calling it from B2 (all the support I've implemented was using CMake). This approach works only if your module doesn't have other modules as dependencies, I guess?

AFAIK the ICE you're hitting with `import std` is really one of the limitations listed [here](https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Modules.html):

> Textual merging of reachable GM entities
>
>    Entities may be multiply defined across different header-units. These must be de-duplicated, and this is implemented across imports, or when an import redefines a textually-defined entity. However the reverse is not implemented—textually redefining an entity that has been defined in an imported header-unit. A redefinition error is emitted.

Except that there's something wrong in the handling they do and it ends up in an ICE. This is quite unfortunate because `import std` makes a huge difference in build speed (the reason why I'm interested in modules), and plays badly with the "compatibility header" approach that Peter and me devised. I narrowed down the ICE to something like

```cpp
import std;
#include <stdlib.h> // ICE
```

While this works fine:


```cpp
#include <stdlib.h>
import std;
```

I still need to report the ICE, since the diagnostic should be "this is not implemented" rather than an ICE, but still.

anarthal:
Still, IMO what you've implemented is likely enough to prove that the support you added for modules is good enough, at least at this point in time. My aim while modularizing is running the entire test suite under modules, but I understand that this might not be feasible. This is much better than what most libraries do.

As a curiosity, this page lists the current status of some libs: https://arewemodulesyet.org/

Once this lib gets accepted into Boost, we can suggest adding it there.

anarthal:
In principle, you need to build the std module yourself for this to work (it's a file located in the libstdc++ include folder). I'd say this should say "no module std" rather than an ICE, but things are prototypy yet.

Issue #1287, opened 2025-12-17
More decimal128_t add sticky bit


mborland:
Since we don't promote to u256 for addition we lose information on the sticky bit of the smaller element. We should be able to make a synthetic routine like umul256 except we return the bool of the sign (if needed)

PR #1288, opened 2025-12-17
Use 256 bit addition and subtraction for 128-bit decimals to avoid loss of precision 

mborland:
Closes: #1287


ckormanyos:
Nice one and nice find, Matt. It's amazing how some of the function tests and awesome client tests really find edge cases

Issue #1311, opened 2026-01-22
Improve square root efficiency


ckormanyos:
The performance of the square root function is not good. The purpose of this issue is to improve it. This issue is partially motivated by discussion in the [ckormanyos/soft_double](https://github.com/ckormanyos/soft_double/issues/129) repo.

The current `sqrt_impl` implementation can be found in `boost/decimal/detail/cmath/sqrt.hpp`. It uses a 2nd-order Padé approximation followed by 3-5 Newton-Raphson iterations, which works but has room for performance improvement.

Attempt to use a table-driven method with higher-order corrections and no iterations.

Cc: @mborland and @justinzhuguangwen

justinzhuguangwen:
Hi @ckormanyos — thanks for the heads up and the Boost 1.91 context. I’ll use #1328 as the new baseline and keep working on the table-driven sqrt when I can. Cheers.

PR #1328, opened 2026-02-02
Rework sqrt cbrt

ckormanyos:
This PR reworks and improves the existing `sqrt` and `cbrt` impls.

- For `sqrt` reduce the argument further to $1/16 < x {\leq} 1/4$ and run the Pade approximation at $1/8$.
- For `cbrt` reduce the argument further to $1/64 < x {\leq} 1/8$ and run the Pade approximation at $1/16$.
- Reduce the loops of Newton iterations on both to $2,3,4$, down from $3,4,5$.
- Decrease the $128$-bit tolerance of `cbrt` to expected value.
- Do some sync-up of the imps `sqrt` and `cbrt`.

This is one that I've had planned for a while. I think it's about as good as I'm going to get it with the current technology.

Note that these calculations are still slow and could _DEFINITELY_ use efficiency improvements, for instance with a better initial guess.

### DIFF
--- a/include/boost/decimal/detail/cmath/cbrt.hpp
+++ b/include/boost/decimal/detail/cmath/cbrt.hpp
@@ -1,4 +1,4 @@
-// Copyright 2023 - 2024 Matt Borland
+// Copyright 2023 - 2026 Matt Borland
 // Copyright 2023 - 2026 Christopher Kormanyos
 // Distributed under the Boost Software License, Version 1.0.
 // https://www.boost.org/LICENSE_1_0.txt
--- a/include/boost/decimal/detail/cmath/sqrt.hpp
+++ b/include/boost/decimal/detail/cmath/sqrt.hpp
@@ -1,4 +1,4 @@
-// Copyright 2023 - 2024 Matt Borland
+// Copyright 2023 - 2026 Matt Borland
 // Copyright 2023 - 2026 Christopher Kormanyos
 // Distributed under the Boost Software License, Version 1.0.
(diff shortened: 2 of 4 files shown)


ckormanyos:
See also #1311

ckormanyos:
Looking at the Boost calendar, we have until the 18th for new libs and another week for larger changes. If this one fits, it might be nice. Let's see if it seems stable.

include/boost/decimal/detail/cmath/cbrt.hpp
```diff
@@ -181,7 +191,7 @@ constexpr auto cbrt_impl(const T x) noexcept
 } //namespace detail
 
 BOOST_DECIMAL_EXPORT template <typename T>
-constexpr auto cbrt(const T val) noexcept
+constexpr auto cbrt(const T val) noexcept // LCOV_EXCL_LINE
```
mborland: One thing I have been thinking on and off about is should we take Intel's approach to implementation when we can? For example they do this:

```c++
auto cbrt(decimal32_t val) noexcept
{
    return static_cast<decimal32_t>(std::cbrt(static_cast<double>(val)));
}
```

`double` exceeds the precision of `decimal32_t`, and an x64 80-bit long double exceeds the precision of `decimal64_t`. Intel assumes the existence of `_Quad` (since it always ships with their compiler) and uses that for `decimal128_t`. 

In order to make everything still `constexpr` we would need to guard these by consteval detection. Thoughts?

PR #1360, opened 2026-04-03
Add CUDA support to decimal128_t

### LINKED ISSUE #1356: Add CUDA support to `decimal128_t` construction and basic operations

mborland:
Closes: #1356


mborland:
These types require an 8k stack to be used. That's pretty big and I'm not sure how valuable this will be.

PR #1365, opened 2026-04-03
Make decimal32_t/decimal64_t/decimal128_t similar to regular int/double with no default initialization

### LINKED ISSUE #1364: Decimal types should act more like normal ints/doubles and shouldn't have default init
this sets it to uint64_t{0}, which isn't true decimal zero `0E-0`, but technically is `0E-398` (which is also different zero representation).
https://github.com/boostorg/decimal/blob/a07d93a465c528bc1e247ad2e2c1d0f2c66cf06e/include/boost/decimal/decimal64_t.hpp#L128

pps83:
closes #1364

### DIFF
--- a/include/boost/decimal/decimal128_t.hpp
+++ b/include/boost/decimal/decimal128_t.hpp
@@ -113,7 +113,7 @@ public:
     using biased_exponent_type = std::int32_t;
 
 private:
-    int128::uint128_t bits_ {};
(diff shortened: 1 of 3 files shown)


pps83:
Note, the fast ones perhaps also should be updated ?

mborland:
Closing since the linked issue has also been closed

Issue #1384, opened 2026-05-02
Internal cmath headers fail to compile when boost::log is in scope (unqualified log resolves to namespace)


acko1980:
## Summary

Several internal `cmath` headers in `boost::decimal` call `log(...)` (and `log1p`, `log10`) unqualified inside template bodies. In any TU where `boost::log` (the Boost.Log namespace, declared in `<boost/log/detail/config.hpp>`) is brought into scope before `<boost/decimal/cmath.hpp>` is parsed, unqualified lookup at template-definition time finds the namespace `boost::log` before `boost::decimal::log` (which is not yet declared at the point those headers run within the umbrella's include order), and the compiler errors out:

- **Clang:** `error: unexpected namespace name 'log': expected expression`
- **GCC:** `error: expected primary-expression before '(' token [-Wtemplate-body]`

Reproduces on both Apple Clang 21.0.0 and GCC 15.2.0; this is standard-conforming behaviour of two-phase template lookup, not a compiler bug.

## Affected headers (v6.0.1)

All call `log` / `log10` / `log1p` unqualified inside template bodies. The seven failing sites match exactly across both Clang and GCC:

| Header | Line |
|---|---|
| `boost/decimal/detail/cmath/acosh.hpp` | 79, 93 |
| `boost/decimal/detail/cmath/asinh.hpp` | 55, 60 |
| `boost/decimal/detail/cmath/atanh.hpp` | 69 |
| `boost/decimal/detail/cmath/impl/ellint_impl.hpp` | 80 |
| `boost/decimal/detail/cmath/log.hpp` | 56 (inside `log_impl`, before `boost::decimal::log` is declared at line 80 of the same header) |

These are all included via the `<boost/decimal/cmath.hpp>` umbrella *before* `detail/cmath/log.hpp` is itself included, so even within the umbrella's own parse, `boost::decimal::log` isn't visible when the dependent calls are first encountered. Two-phase lookup defers ADL on dependent arguments to instantiation, but the unqualified function-name lookup at template-*definition* time succeeds against `boost::log` (a namespace), which is invalid in expression context, and parsing fails.

## Reproducer

```cpp
// repro.cpp
#include <boost/log/trivial.hpp>     // brings boost::log namespace into scope
#include <boost/decimal/cmath.hpp>   // fails to parse

int main() { return 0; }
```

### Verified on

**Apple Clang 21.0.0** (`clang-2100.0.123.102`, arm64-apple-darwin25.4.0):

```
[...]

PR #1385, opened 2026-05-02
Potential proposal for 1384

ckormanyos:
Here is a potential proposal to solve #1384. It uses fowrward declarations only.

I think it would still be wise to combine these forward declarations additionally with the consistent use of namespace-decorated log-functions in the internal headers of Decimal's `cmath`.

Cc: @mborland

### DIFF
--- a/include/boost/decimal/cmath.hpp
+++ b/include/boost/decimal/cmath.hpp
@@ -5,6 +5,7 @@
 #ifndef BOOST_DECIMAL_CMATH_HPP
 #define BOOST_DECIMAL_CMATH_HPP
 
+#include <boost/decimal/detail/fwd_log.hpp>
 #include <boost/decimal/detail/promotion.hpp>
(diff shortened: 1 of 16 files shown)


ckormanyos:
Hi Matt (@mborland) I'm satisfied with this resolution of the `boost::log` namespace conflict (at least for now). If you could take a quick glance and see if you're OK with it, then it's good to go.

mborland:
Let me actually pull in Boost.Log as part of our test suite for completeness. I'll just add a commit or two onto this branch. Otherwise looks generally good.

mborland (approved):
Once this cycles green I'll merge it in

mborland:
I'm reverting my commits to your green one. Boost.log and filesystem have too many problems to possibly compile under -Werror. We'd have to fix upstream since it's in the actual compilation steps of the library.

PR #1389, opened 2026-05-06
Fix cohort handling of zero to match IEEE 754 requirements

### LINKED ISSUE #1387: Remove canonicalization of 0
Per IEEE 754-2008 section 3.5.1: "A zero has a much larger cohort: the cohort of +0 contains a representation for each exponent, as does the cohort of −0". This is easy to fix in the constructor, but will affect basically every operation downstream.

mborland:
Closes: #1387


mborland:
These 32bit MSVC 14.5 failures are bizarre. It loses the sign and has stack overflows all over the place where the other MSVC versions have been green for a while. I'm going to move adding that runner to a different PR so that I can investigate what's going on better. These changes are still the correct ones.

PR #1391, opened 2026-05-08
Add MSVC 14.5 runner

mborland:


mborland:
Over alignment was the cause of our decimal type issues. Will have to dig into big uints on Monday. I believe that I have forced over alignment with u256.

PR #1397, opened 2026-06-12
Docs: edit_url setting

sdarwin:
After feedback from Ruben (boostorg/redis) the `edit_url` setting seems more appropriate than `branches`. It is exactly aimed at the issue.


mborland:
Thanks Sam

PR #1423, opened 2026-08-12
changes for JOSS publication

danielskatz:


mborland (approved):
These all look good to me. Thank you!

Issue #1429, opened 2026-08-27
C23 <math.h> Functions


ibmibmibm:
[P3935R0 "Rebasing `<cmath>` on C23"](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p3935r0.html)
proposes pulling the remaining C23 `<math.h>` functions into C++, since C++26 is based on C23
but `<cmath>` is still based on C17. The paper puts decimal floating-point out of its own scope,
but nearly everything it lists is an ISO/IEC 60559 operation that
[C23 Annex H](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3220.pdf) already requires for
`_Decimal32` / `_Decimal64` / `_Decimal128`. Boost.Decimal should provide all of them regardless
of what C++29 adopts — the paper even notes that `iscanonical` and `canonicalize` are "primarily
motivated by ISO/IEC 60559 decimal floating-point representations."

This issue tracks that surface across all six types (`decimal32_t`, `decimal64_t`, `decimal128_t`,
`decimal_fast32_t`, `decimal_fast64_t`, `decimal_fast128_t`), plus the decimal-only functions from
C23 Annex H that P3935R0 does not cover.

Checked items are already provided; a trailing note gives our current spelling where it differs
from C23.

## Classification and canonicalization

- [x] `issignaling` — provided, documented under "Non-standard Functions"
- [ ] `iscanonical` — C23 specifies exactly which decimal encodings are canonical, so unlike the
      binary case this has a hard definition (non-canonical declets in DPD, significands above
      the format maximum in BID)
- [ ] `issubnormal`
- [ ] `iszero`
- [ ] `canonicalize`

## Comparison

- [ ] `iseqsig` — `compareSignalingEqual`; raises `FE_INVALID` on quiet NaN, unlike `operator==`
- [ ] `totalorder` — could alias the existing `comparetotal`; see open questions
- [ ] `totalordermag`

## NaN payload

- [ ] `getpayload` — could alias the existing `read_payload`
- [ ] `setpayload`
- [ ] `setpayloadsig`

## Trigonometric (pi-scaled)

- [ ] `acospi`
- [ ] `asinpi`
- [ ] `atanpi`
- [ ] `atan2pi`
- [ ] `cospi`
- [ ] `sinpi`
- [ ] `tanpi`

## Exponential and logarithmic

- [ ] `exp10`
- [ ] `exp10m1`
- [ ] `exp2m1`
[...]

PR #1431, opened 2026-08-27
Add roundeven

ibmibmibm:
Fix roundeven in #1429


mborland (approved):
Seems good to me. Will merge once CI finishes cycling. The fix for the Drone failure is in #1434

PR #1432, opened 2026-08-28
Correct the range of the log1p polynomial

### LINKED ISSUE #1430: The log1p function gives no correct digits near x = -1, and loses digits above |x| = 0.14
## Summary

The `boost::decimal::log1p(x)` function calculates a Taylor polynomial of
`log(1 + x)`. It uses this polynomial for all values of `x` in the range `(-1, 0.5]`.
But the coefficient tables are not long enough for this range. The code has an upper
limit at `x = 0.5`. It has no lower limit.

The function does not report a failure. It gives a result that does not have enough
correct digits.

At `x = 0.5`, the `decimal128_t` type gives only 12 correct digits of 34. At the
same point, the `decimal64_t` type gives only 7 correct digits of 16. The error becomes
larger when `|x|` becomes larger. At `x = -0.9`, the `decimal128_t` type gives
only 2 correct digits. This value of `x` is inside the range of the branch.

Near `x = -1`, the result has no correct digits. At `x = -0.99`, the relative
error is 16 percent for the `decimal128_t` type. It is 25 percent for the `decimal64_t`
type, and 34 percent for the `decimal32_t` type. At `x = -0.999`, the same three errors
are 40 percent, 48 percent and 54 percent. The first digit is already incorrect.

All six decimal types have this problem. The `decimal_fast*` types have their own copies
of the coefficients, but the values are the same. The errors are also the same.

## Environment

- Boost.Decimal at commit `9cbdcf0d` of `develop`
- The correct values come from MPFR 4.2.2. The tests call `mpfr_log1p` with 300 bits. A
  `decimal128_t` value has 34 digits, which needs 113 bits.
- A second check uses Boost.Math 1.92 with `boost::math::log1p<cpp_bin_float_50>`.

## Test program

This program needs Boost.Decimal and MPFR. It does not contain the correct values. It
calculates them with MPFR when it runs. To build it, use
`g++ -std=c++17 -O2 log1p_test.cpp -o log1p_test -lmpfr`.

The program tests 20 values of `x` in the branch. For each value, it calls `log1p(x)`
and `log(1 + x)`. It compares the two results with `mpfr_log1p`. The table gives the
value of `log1p(x)`, the correct value, and the error of each result. The unit of the
error is eps = `10^(1-digits10)`.

The output of the program does not change between 150 bits and 1000 bits in MPFR.

<details>

```cpp
#include <boost/decimal.hpp>
#include <mpfr.h>
#include <cstdio>
#include <cstring>
#include <limits>

static const char* XS[] = {
    "-0.999", "-0.99", "-0.9", "-0.75", "-0.5", "-0.45", "-0.4", "-0.3", "-0.25",
    "-0.2", "-0.1", "-0.05", "0.05", "0.1", "0.2", "0.25", "0.3", "0.4", "0.45", "0.5"};

// Move a decimal value into MPFR. A decimal value has an exact decimal string.
template <typename T>
static void put(mpfr_t m, T v, char* s)
{
    auto r = to_chars(s, s + 150, v, chars_format::scientific, std::numeric_limits<T>::digits10 - 1);
    *r.ptr = '\0';
    mpfr_set_str(m, s, 10, MPFR_RNDN);
}

// Relative error in units of eps, where eps = 10^(1-digits10).
static double err_eps(mpfr_srcptr got, mpfr_srcptr ref, int d)
{
    mpfr_t e, p;
    mpfr_inits2(300, e, p, (mpfr_ptr) 0);
    mpfr_sub(e, got, ref, MPFR_RNDN);
    mpfr_div(e, e, ref, MPFR_RNDN);
    mpfr_abs(e, e, MPFR_RNDN);
    mpfr_ui_pow_ui(p, 10, static_cast<unsigned long>(d - 1), MPFR_RNDN);
    mpfr_mul(e, e, p, MPFR_RNDN);
    const double v = mpfr_get_d(e, MPFR_RNDN);
    mpfr_clears(e, p, (mpfr_ptr) 0);
    return v;
}

template <typename T>
static void run(const char* name)
{
    constexpr int d = std::numeric_limits<T>::digits10;
    const int w = d + 7;
    std::printf("%s (%d digits)\n\n%6s  %*s  %*s  %9s %9s\n", name, d, "x",
                w, "log1p(x)", w, "correct (MPFR)", "log1p err", "log err");

    for (const char* xs : XS)
    {
        T x {};
        from_chars(xs, xs + std::strlen(xs), x);
        const T a = log1p(x);
        const T b = log(x + T {1});

        mpfr_t xm, ref, am, bm;
        mpfr_inits2(300, xm, ref, am, bm, (mpfr_ptr) 0);
        char sx[160], sa[160], sb[160], sr[160];
        put(xm, x, sx);
        mpfr_log1p(ref, xm, MPFR_RNDN);
        put(am, a, sa);
        put(bm, b, sb);
        mpfr_snprintf(sr, sizeof sr, "%.*Re", d - 1, ref);

        std::printf("%6s  %*s  %*s  %9.2e %9.2e\n", xs, w, sa, w, sr,
                    err_eps(am, ref, d), err_eps(bm, ref, d));
        mpfr_clears(xm, ref, am, bm, (mpfr_ptr) 0);
    }
}

int main()
{
    run<decimal32_t>("decimal32_t");
    std::printf("\n");
    run<decimal64_t>("decimal64_t");
    std::printf("\n");
    run<decimal128_t>("decimal128_t");
    return 0;
}
```

</details>

### Output of the program

<details>
<summary>Show the output</summary>

```
decimal32_t (7 digits)

     x        log1p(x)  correct (MPFR)  log1p err   log err
-0.999   -3.167173e+00   -6.907755e+00   5.42e+05  4.04e-02
 -0.99   -3.053940e+00   -4.605170e+00   3.37e+05  4.04e-02
  -0.9   -2.190365e+00   -2.302585e+00   4.87e+04  4.04e-02
 -0.75   -1.381965e+00   -1.386294e+00   3.12e+03  2.60e-01
  -0.5   -6.931390e-01   -6.931472e-01   1.18e+01  2.80e-02
 -0.45   -5.978354e-01   -5.978370e-01   2.68e+00  1.26e-03
  -0.4   -5.108252e-01   -5.108256e-01   8.30e-01  5.41e-01
  -0.3   -3.566748e-01   -3.566749e-01   4.04e-01  4.04e-01
 -0.25   -2.876820e-01   -2.876821e-01   2.52e-01  2.69e+00
  -0.2   -2.231436e-01   -2.231436e-01   2.18e-01  2.30e-01
  -0.1   -1.053605e-01   -1.053605e-01   1.49e-01  2.05e+00
 -0.05   -5.129330e-02   -5.129329e-02   1.09e-01  9.27e+00
  0.05    4.879016e-02    4.879016e-02   8.55e-02  8.55e-02
   0.1    9.531018e-02    9.531018e-02   2.05e-03  3.17e-01
   0.2    1.823216e-01    1.823216e-01   2.37e-01  8.60e-01
  0.25    2.231436e-01    2.231436e-01   2.18e-01  2.30e-01
   0.3    2.623643e-01    2.623643e-01   1.35e-01  6.27e-01
   0.4    3.364724e-01    3.364722e-01   4.86e-01  4.06e-01
  0.45    3.715643e-01    3.715636e-01   2.00e+00  1.52e-01
   0.5    4.054681e-01    4.054651e-01   7.38e+00  2.27e-01

decimal64_t (16 digits)

     x                 log1p(x)           correct (MPFR)  log1p err   log err
-0.999   -3.624463262921586e+00   -6.907755278982137e+00   4.75e+14  1.37e-01
 -0.99   -3.445429935829793e+00   -4.605170185988091e+00   2.52e+14  1.37e-01
  -0.9   -2.268560161700878e+00   -2.302585092994046e+00   1.48e+13  1.37e-01
 -0.75   -1.386004416075499e+00   -1.386294361119891e+00   2.09e+11  2.75e-01
  -0.5   -6.931471597575605e-01   -6.931471805599453e-01   3.00e+07  1.31e-01
 -0.45   -5.978369988799574e-01   -5.978370007556204e-01   3.14e+06  8.26e-02
  -0.4   -5.108256236363804e-01   -5.108256237659907e-01   2.54e+05  8.16e-01
  -0.3   -3.566749439385324e-01   -3.566749439387324e-01   5.61e+02  3.39e-01
 -0.25   -2.876820724517775e-01   -2.876820724517809e-01   1.19e+01  1.83e+00
  -0.2   -2.231435513142098e-01   -2.231435513142098e-01   1.98e-01  1.98e-01
  -0.1   -1.053605156578263e-01   -1.053605156578263e-01   1.17e-02  1.17e-02
 -0.05   -5.129329438755055e-02   -5.129329438755053e-02   3.23e-01  1.04e+01
  0.05    4.879016416943200e-02    4.879016416943200e-02   6.28e-02  3.34e+00
   0.1    9.531017980432486e-02    9.531017980432486e-02   4.61e-04  8.40e-01
   0.2    1.823215567939546e-01    1.823215567939546e-01   1.44e-01  1.44e-01
  0.25    2.231435513142118e-01    2.231435513142098e-01   9.16e+00  1.98e-01
   0.3    2.623642644676019e-01    2.623642644674911e-01   4.22e+02  5.64e-01
   0.4    3.364722366790574e-01    3.364722366212129e-01   1.72e+05  9.07e-02
  0.45    3.715635571786926e-01    3.715635564324830e-01   2.01e+06  1.78e-01
   0.5    4.054651154407084e-01    4.054651081081644e-01   1.81e+07  2.02e-01

decimal128_t (34 digits)

     x                                   log1p(x)                             correct (MPFR)  log1p err   log err
-0.999   -4.164916650246122732995194259803316e+00   -6.907755278982137052053974364053093e+00   3.97e+32  9.02e-02
 -0.99   -3.862452991606967326819408602026066e+00   -4.605170185988091368035982909368728e+00   1.61e+32  9.02e-02
  -0.9   -2.298577268642851827874575201361194e+00   -2.302585092994045684017991454684364e+00   1.74e+30  9.02e-02
 -0.75   -1.386292605981115997609940521878940e+00   -1.386294361119890618834464242916353e+00   1.27e+27  9.82e-02
  -0.5   -6.931471805597585173503552917804030e-01   -6.931471805599453094172321214581766e-01   2.69e+20  1.90e-01
 -0.45   -5.978370007556173370791674450059346e-01   -5.978370007556204493732799981774115e-01   5.21e+18  4.61e-01
  -0.4   -5.108256237659906506145698237257048e-01   -5.108256237659906832055140963036619e-01   6.38e+16  1.50e+00
  -0.3   -3.566749439387323789121363369167257e-01   -3.566749439387323789126387112411845e-01   1.41e+12  1.18e+00
 -0.25   -2.876820724517809274392185455162198e-01   -2.876820724517809274392190059938274e-01   1.60e+09  1.15e+00
  -0.2   -2.231435513142097557662950902199866e-01   -2.231435513142097557662950903098345e-01   4.03e+05  8.81e-01
  -0.1   -1.053605156578263012275009808393128e-01   -1.053605156578263012275009808393128e-01   1.61e-02  1.61e-02
 -0.05   -5.129329438755053342619614425468725e-02   -5.129329438755053342619614425468724e-02   2.25e-01  5.04e+00
  0.05    4.879016416943200306537440422316466e-02    4.879016416943200306537440422316466e-02   2.85e-02  1.05e+00
   0.1    9.531017980432486004395212328076509e-02    9.531017980432486004395212328076509e-02   2.33e-02  1.34e+00
   0.2    1.823215567939546262117180252150547e-01    1.823215567939546262117180251545146e-01   3.32e+05  3.66e-01
  0.25    2.231435513142097557662953703824633e-01    2.231435513142097557662950903098345e-01   1.26e+09  8.81e-01
   0.3    2.623642644674910520357710756957145e-01    2.623642644674910520354959868809544e-01   1.05e+12  3.92e-01
   0.4    3.364722366212129448127661242093529e-01    3.364722366212129305045934102169921e-01   4.25e+16  1.22e+00
  0.45    3.715635564324842483126398844816274e-01    3.715635564324830337480484562193708e-01   3.27e+18  1.89e-01
   0.5    4.054651081082287609943461159105443e-01    4.054651081081643819780131154643491e-01   1.59e+20  9.02e-02
```

</details>

The error of `log1p` becomes very large when `|x|` becomes larger. At `x = 0.5`, the
error is 1.59e+20 eps for the `decimal128_t` type, and 1.81e+07 eps for the
`decimal64_t` type. At `x = -0.999`, the error is 3.97e+32 eps for the `decimal128_t`
type, and 5.42e+05 eps for the `decimal32_t` type.

The error of `log(1 + x)` is less than 11 eps at all 20 points. This shows that the
`log` function is correct. Only the series branch of the `log1p` function is incorrect.

But do not use `log(1 + x)` as the correction for all values. Its error becomes larger
near `x = 0`. At `x = -0.05`, the error is 9.27 eps for the `decimal32_t` type. The
section "Why log(1 + x) is not the correction for all values" gives more data.

To test the `decimal_fast32_t`, `decimal_fast64_t` and `decimal_fast128_t` types, change
the three calls in `main`. The program then gives the same 20 rows, with the same
errors.

## Cause

The `log1p_impl` function in `include/boost/decimal/detail/cmath/log1p.hpp` selects the
branch:

<details>
<summary>Show the table in the library</summary>

```cpp
if (x > T { 5, -1 })
{
    result = ::boost::decimal::log(x + one);
}
else
{
    result = x * fma(detail::log1p_series_expansion(x), x, one);
}
```

</details>

The code uses the polynomial for all values of `x` from `-1` to `0.5`. This test has a
limit at `0.5` for positive values. It has no limit for negative values.

The tables in `include/boost/decimal/detail/cmath/impl/log1p_impl.hpp` contain the
Taylor coefficients of `log(1 + x)`:

| table | number of terms | highest term | truncation error at \|x\| = 0.5 |
| --- | --- | --- | --- |
| `d32_coeffs` | 12 | `x^13` | `0.5^14/14` = 4.4e-06 |
| `d64_coeffs` | 20 | `x^21` | `0.5^22/22` = 1.1e-08 |
| `d128_coeffs` | 36 | `x^37` | `0.5^38/38` = 9.6e-14 |

The errors in this table are almost the same as the errors from the test. This shows the
cause. The code uses the tables outside their correct range.

### The correct range of each type

This table shows the first value of `|x|` where the error is more than 1 eps. The unit
eps is `10^(1-digits10)`. The test uses steps of 0.001.

| type | highest term | error starts above x = | error starts below x = |
| --- | --- | --- | --- |
| `decimal32_t` | `x^13` | 0.426 | -0.402 |
| `decimal64_t` | `x^21` | 0.226 | -0.217 |
| `decimal128_t` | `x^37` | 0.143 | -0.142 |

The branch has a width of 1.5, from `-1` to `0.5`. The `decimal128_t` type is correct
for a width of 0.285, which is 19 percent of the branch. The `decimal64_t` type is
correct for 30 percent, and the `decimal32_t` type for 55 percent.

## Other functions with the same problem

Three other functions call `log1p`: `atanh`, `asinh` and `acosh`. These are the only
calls to `log1p` in the library. Each one uses `log1p` near zero, because `log1p` is
correct near zero. But each one sends a value to `log1p` that is larger than its own
argument. This value goes past the correct limit of the polynomial, and the error goes
into the result.

These are the calls, with the range of each branch:

<details>
<summary>Show the three calls</summary>

```cpp
atanh(x):  log1p(x) - log1p(-x)                fourth_root_eps <= |x| < 0.5
asinh(x):  log1p(x + (sqrt(1 + x*x) - 1))      fourth_root_eps <=  x  < 0.5
acosh(x):  log1p(y + sqrt(y*y + 2*y))          y = x - 1, root_eps <= y, x < 1.5
```

</details>

The next table gives the range of `x` that loses digits. In this range, the value that
reaches `log1p` is more than the correct limit of the polynomial. But it is not more
than 0.5.

| function | `decimal32_t` | `decimal64_t` | `decimal128_t` |
| --- | --- | --- | --- |
| `atanh` | 0.375 to 0.500 | 0.200 to 0.500 | 0.125 to 0.500 |
| `asinh` | 0.324 to 0.417 | 0.183 to 0.417 | 0.118 to 0.417 |
| `acosh` | 1.051 to 1.083 | 1.017 to 1.083 | 1.007 to 1.083 |

The errors from the test show these ranges. This table gives the error in units of eps.
The column `log1p argument` gives the value that reaches `log1p`.

| function | x | log1p argument | `decimal32_t` | `decimal64_t` | `decimal128_t` |
| --- | --- | --- | --- | --- | --- |
| `atanh` | 0.20 | 0.2000 | 2.27e-01 | 4.44e-02 | 7.23e+04 |
| `atanh` | 0.30 | 0.3000 | 1.36e-02 | 1.44e+02 | 3.67e+11 |
| `atanh` | 0.40 | 0.4000 | 3.07e-01 | 8.47e+04 | 2.16e+16 |
| `atanh` | 0.49 | 0.4900 | 3.43e+00 | 7.80e+06 | 5.13e+19 |
| `atanh` | 0.50 | 0.5000 | 2.63e-01 | 2.81e-01 | 2.69e-01 |
| `asinh` | 0.20 | 0.2198 | 3.58e+00 | 3.56e+00 | 1.08e+07 |
| `asinh` | 0.30 | 0.3440 | 1.51e+00 | 7.39e+03 | 1.64e+14 |
| `asinh` | 0.40 | 0.4770 | 2.26e+00 | 6.78e+06 | 2.81e+19 |
| `asinh` | 0.45 | 0.5466 | 3.87e-01 | 7.49e-01 | 6.90e-01 |
| `acosh` | 1.01 | 0.1518 | 4.92e-01 | 5.47e-01 | 1.24e+01 |
| `acosh` | 1.02 | 0.2210 | 2.89e-01 | 8.68e-01 | 1.32e+07 |
| `acosh` | 1.05 | 0.3702 | 1.80e-01 | 3.41e+04 | 2.44e+15 |
| `acosh` | 1.08 | 0.4879 | 5.23e+00 | 1.09e+07 | 6.45e+19 |
| `acosh` | 1.10 | 0.5583 | 3.48e-01 | 4.26e-01 | 9.32e-01 |

The error becomes small again at the end of each range. At that point the value that
reaches `log1p` becomes more than 0.5, and `log1p` uses `log(x + 1)`. This makes a large
step in the error. The step does not come from `atanh`, `asinh` or `acosh`. It is at
the branch limit of `log1p`.

The last row of each function shows this step. For the `decimal128_t` type, the error of
`acosh` decreases from 6.45e19 eps at `x = 1.08` to 0.93 eps at `x = 1.10`.

The `decimal32_t` type loses only a few eps, because its polynomial is correct up to
0.375. The `decimal64_t` and `decimal128_t` types lose almost all their digits.

To measure these errors, change two lines of the test program above. Call the function
in place of `log1p`, and call `mpfr_atanh`, `mpfr_asinh` or `mpfr_acosh` in place of
`mpfr_log1p`.

The `legendre.hpp` file also includes `log1p.hpp`, but the `legendre` function does not
call `log1p`. It uses only a recurrence with add, subtract, multiply and divide. A test
compares it with `boost::math::legendre_p<cpp_bin_float_50>` at 14 values of `x` and 8
orders. The largest error is 34.8 eps for `decimal32_t`, 19.7 eps for `decimal64_t` and
1.91 eps for `decimal128_t`.

The error becomes smaller when the precision becomes larger. For the three functions
above, the error becomes larger. The largest error is at `n = 13` and `x = 0.45`, where
`P_13(0.45)` is 0.00524. This value is near a root, so the recurrence loses digits
there. The `legendre.hpp` file does not need to include `log1p.hpp`.

A correction to `log1p` also corrects these three functions. They need no change.

## Possible corrections

A correction needs two changes.

The first change is a lower limit for the series branch. The code has an upper limit at
`x = 0.5`, but it has no lower limit. Values of `x` near `-1` go to the polynomial.
A polynomial cannot give a correct result there. When `x` is near `-1`, the value of
`log1p(x)` is very large and negative.

The second change is a better polynomial. Use `log(x + 1)` for all values outside
the range of the polynomial.

The sections below give three possible methods. The tests measure all three.

### Option 1: keep the coefficients, but make the range smaller

Use the tables only for values of `|x|` inside their correct limit. The limits are
0.375, 0.200 and 0.125. Use `log(x + 1)` outside these limits.

This is the change. The limit function is new in `log1p_impl.hpp`.

<details>
<summary>Show the new limit function</summary>

```cpp
template <typename T>
constexpr auto log1p_series_limit() noexcept -> T
{
    return std::numeric_limits<T>::digits10 <= 7  ? T { 375, -3 }
         : std::numeric_limits<T>::digits10 <= 16 ? T { 200, -3 }
         :                                          T { 125, -3 };
}
```

</details>

In `log1p_impl`, use this limit in place of the test `x > T { 5, -1 }`:

<details>
<summary>Show the change in log1p_impl</summary>

```cpp
if (abs(x) > detail::log1p_series_limit<T>())
{
    result = ::boost::decimal::log(x + one);
}
else
{
    result = x * fma(detail::log1p_series_expansion(x), x, one);
}
```

</details>

The test `abs(x)` also gives the lower limit. The old test has no lower limit.

### Option 2: change the variable

The comment on `d128_coeffs` shows this form:

```
log1p(x) = 2 * atanh(w),   w = x / (2 + x)
```

The coefficients are `2/(2k+1)`, for `k = 0, 1, 2, ...`. The code calculates
`w * P(w * w)`, where `P` is the polynomial. For values of `|x|` less than or
equal to 0.5, the value of `|w|` is less than or equal to 1/3. The value `-0.5` gives
the largest `|w|`. This value sets the number of terms. Use `log(x + 1)` for `|x|`
more than 0.5.

Calculate this polynomial inside `log1p`. Do not call `boost::decimal::atanh`. The
`atanh` function calls `log1p`, so a call from `log1p` to `atanh` makes a loop. The
loop does stop, because `|w|` becomes about one half at each step, and `atanh` uses its
own Taylor series below `fourth_root_epsilon`. But `atanh` makes two calls to `log1p`,
so the number of calls becomes two times larger at each step. For the `decimal128_t`
type this gives about 3e7 calls for one result.

This is the change. The coefficients are `2/(2k+1)`. This function shows the values. The
correction puts them in a table of constant values, in the same form as `d64_coeffs`.
The section "Number of terms" gives the count `N` for each type. The name `coeffs` below
is this table.

<details>
<summary>Show the coefficient function</summary>

```cpp
template <typename T, std::size_t N>
constexpr auto make_atanh_coeffs() noexcept -> std::array<T, N>
{
    std::array<T, N> a {};
    for (std::size_t k {}; k < N; ++k) { a[k] = T { 2, 0 } / T(2 * k + 1); }
    return a;
}
```

</details>

In `log1p_impl`, replace the two branches inside the last `else`:

<details>
<summary>Show the change in log1p_impl</summary>

```cpp
if (abs(x) > T { 5, -1 })
{
    result = ::boost::decimal::log(x + one);
}
else
{
    const T w { x / (T { 2, 0 } + x) };
    result = w * detail::taylor_series_result(w * w, coeffs);
}
```

</details>

### Option 3: the method of glibc

The file `sysdeps/ieee754/dbl-64/s_log1p.c` in glibc uses the same change of variable,
but it first changes the argument to a small value. The method has these steps:

1. If `-0.29289 < x < 0.41421`, set `k = 0`. Set `f = x`. This step does not change
   `x`, and the value `f` is exact.
2. For all other values of `x`, calculate `u = 1 + x`. Calculate the exact error `c` of
   this sum. Find `k` and `f` for the equation `u = 2^k * (1 + f)`. The value `1 + f`
   must be between `sqrt(2)/2` and `sqrt(2)`.
3. Calculate `s = f / (2 + f)`. Calculate `z = s * s`. The value of `|s|` is now less
   than 0.1716.
4. Calculate `R = Lp1*z + Lp2*z^2 + ...`. Calculate `hfsq = f*f/2`.
5. Calculate the result: `k*ln2 + (f - (hfsq - (s*(hfsq + R) + c)))`.

The small range of `s` is important. This method needs fewer terms than option 2.

glibc uses Remez coefficients for the `double` type. The array `Lp` in
`sysdeps/ieee754/dbl-64/s_log1p.c` contains 7 of these values. The test below does not
use Remez coefficients. It uses the Taylor coefficients `2/(2k+1)`, which are the same
coefficients as option 2. All the results below come from these Taylor coefficients. The
term counts below are an upper limit.

Remez coefficients decrease the counts more. The section "The Remez program" below
gives a program that calculates them. It is a weighted exchange algorithm in the
`Decimal` type of Python, with 120 digits. The weight makes the error of the polynomial
equal to the relative error of the result.

The program gives the same count as glibc. For the `double` type it gives 7 terms, with
an error of 1.2e-18. glibc also uses 7 terms. The tables below give the counts for the
decimal types.

The code must calculate the value `k*ln2` to the full precision of the type. glibc keeps
`ln2` in two parts, `ln2_hi` and `ln2_lo`. The value `k*ln2_hi` is then exact. The test
below does the same.

This is the change. It replaces the last `else` of `log1p_impl`. It needs
`frexp.hpp`, and `numbers.hpp` for `sqrt2_v`. The `log1p.hpp` file already includes
`numbers.hpp`. The name `coeffs` below is the table from `make_r_coeffs`, with the
count `N` from the section "Number of terms".

<details>
<summary>Show the new tables</summary>

```cpp
// ln2 in two parts, so that k * ln2_hi is exact. glibc does the same.
template <typename T> struct ln2_split;

template <> struct ln2_split<decimal32_t>
{
    static constexpr decimal32_t hi { 693, -3 };
    static constexpr decimal32_t lo { UINT64_C(1471805599453094), -19 };
};

template <> struct ln2_split<decimal64_t>
{
    static constexpr decimal64_t hi { UINT64_C(69314718), -8 };
    static constexpr decimal64_t lo { UINT64_C(5599453094172321), -25 };
};

template <> struct ln2_split<decimal128_t>
{
    static constexpr decimal128_t hi { UINT64_C(69314718055994531), -17 };
    static constexpr decimal128_t lo
        { -decimal128_t { UINT64_C(5827678785418234), -34 } };
};

// R coefficients: 2/(2k+3) for k = 0, 1, 2, ...  These are glibc's Lp1 to Lp7.
template <typename T, std::size_t N>
constexpr auto make_r_coeffs() noexcept -> std::array<T, N>
{
    std::array<T, N> a {};
    for (std::size_t k {}; k < N; ++k) { a[k] = T { 2, 0 } / T(2 * k + 3); }
    return a;
}
```

</details>

<details>
<summary>Show the change in log1p_impl</summary>

```cpp
constexpr T two { 2, 0 };

T f {};
T c { 0, 0 };
int k { 0 };

// 1 + x inside (sqrt(2)/2, sqrt(2)) needs no reduction, and f = x is exact.
if (x > T { -292893218813452, -15 } && x < T { 414213562373095, -15 })
{
    f = x;
}
else
{
    const T u { one + x };
    c = ((u >= one) ? (one - (u - x)) : (x - (u - one))) / u;  // exact error of the sum

    int e {};
    T m { frexp(u, &e) };                    // u = m * 2^e, m inside [0.5, 1)
    T mm { m + m };
    k = e - 1;
    if (mm > numbers::sqrt2_v<T>) { mm /= two; ++k; }
    f = mm - one;                            // 1 + f inside [sqrt(2)/2, sqrt(2))
}

const T hfsq { (f * f) / two };
const T s { f / (two + f) };                 // |s| <= 0.1716
const T z { s * s };
const T r { z * detail::taylor_series_result(z, coeffs) };

if (k == 0)
{
    result = f - (hfsq - (s * (hfsq + r) + c));
}
else
{
    const T kd { k };
    result = kd * ln2_split<T>::hi
           + (f - (hfsq - (s * (hfsq + r) + (kd * ln2_split<T>::lo + c))));
}
```

</details>

The value `c` is the error of the sum `1 + x`. Option 3 needs this correction, because
step 2 uses `1 + x` and not `x`. Without `c`, the result loses the digits that the sum
loses.

### The Remez program

This program calculates the Remez coefficients. It needs only the standard library of
Python. The command `python3 remez.py opt2 1 22` gives the counts for option 2. The
command `python3 remez.py opt3 1 15` gives the counts for option 3. The program writes
each set of coefficients to a file.

<details>

```python
"""Discrete weighted Remez exchange for the log1p atanh polynomials.

option 2 (no reduction):    2*atanh(w) = w * P(z), z = w*w, |w| <= 1/3
    exact P is g2(z) = sum_k 2 z^k/(2k+1);  rel. err of result = |dP|/g2  -> weight 1/g2
option 3 (glibc reduction): R = z * P(z), result ~ 2*atanh(s), |s| <= (sqrt2-1)/(sqrt2+1)
    exact P is g3(z) = sum_k 2 z^k/(2k+3);  rel. err of result = z|dP|/g2 -> weight z/g2
"""
import math, sys
from decimal import Decimal as D, getcontext
getcontext().prec = 120
TINY = D(10) ** -110

def series(z, off):
    """Return sum_k 2 z^k / (2k+off)."""
    s, zp = D(0), D(1)
    for k in range(800):
        t = zp / D(2 * k + off)
        s += t
        if k and t < TINY:
            break
        zp *= z
    return s * 2

def solve(A):
    """Solve the augmented system A by gaussian elimination with partial pivot."""
    n = len(A)
    for c in range(n):
        p = max(range(c, n), key=lambda r: abs(A[r][c]))
        A[c], A[p] = A[p], A[c]
        pv = A[c][c]
        for r in range(n):
            if r == c: continue
            f = A[r][c] / pv
            if f:
                for k in range(c, n + 1):
                    A[r][k] -= f * A[c][k]
    return [A[i][n] / A[i][i] for i in range(n)]

def poly(c, z):
    v = D(0)
    for a in reversed(c):
        v = v * z + a
    return v

class Grid:
    """Fixed sample grid with cached g and w."""
    def __init__(self, zmax, off, weight_z, M=6000):
        self.z = [zmax * D(i) / M for i in range(1, M + 1)]
        self.g = [series(t, off) for t in self.z]
        gg = self.g if off == 1 else [series(t, 1) for t in self.z]
        self.w = [(t / q if weight_z else D(1) / q) for t, q in zip(self.z, gg)]
        self.M = M

def errs(G, c):
    return [w * (poly(c, z) - g) for z, g, w in zip(G.z, G.g, G.w)]

def alternation(e):
    ext = []
    for i in range(len(e)):
        a = e[i - 1] if i else D(0)
        b = e[i + 1] if i + 1 < len(e) else a
        if (e[i] >= a and e[i] >= b) or (e[i] <= a and e[i] <= b):
            if ext and (e[i] > 0) == (e[ext[-1]] > 0):
                if abs(e[i]) > abs(e[ext[-1]]): ext[-1] = i
            else:
                ext.append(i)
    return ext

def remez(G, N, iters=40):
    idx = [min(G.M - 1, max(0, int(G.M * (1 - math.cos(math.pi * (i + 0.5) / (N + 1))) / 2)))
           for i in range(N + 1)]
    best = None
    for _ in range(iters):
        A = []
        for r, i in enumerate(idx):
            z, g, w = G.z[i], G.g[i], G.w[i]
            A.append([z ** j for j in range(N)] + [D(-1 if r % 2 else 1) / w, g])
        sol = solve(A)
        c, E = sol[:N], abs(sol[N])
        e = errs(G, c)
        ext = alternation(e)
        mx = max(abs(e[i]) for i in ext)
        best = (c, mx)
        if E and abs(mx - E) / mx < D("1e-9"):
            break
        while len(ext) > N + 1:
            ext.pop(0 if abs(e[ext[0]]) < abs(e[ext[-1]]) else -1)
        if len(ext) < N + 1:
            break
        idx = ext
    return best

def taylor(G, N, off):
    c = [D(2) / D(2 * k + off) for k in range(N)]
    return max(abs(x) for x in errs(G, c))

SQRT2 = D(2).sqrt()
CASES = {
    "opt3": (((SQRT2 - 1) / (SQRT2 + 1)) ** 2, 3, True),
    "opt2": (D(1) / 9, 1, False),
}

if __name__ == "__main__":
    case, lo, hi = sys.argv[1], int(sys.argv[2]), int(sys.argv[3])
    zmax, off, wz = CASES[case]
    G = Grid(zmax, off, wz)
    print("=== %s  zmax=%.8f ===" % (case, zmax), flush=True)
    print("   N        remez           taylor", flush=True)
    for N in range(lo, hi + 1):
        c, e = remez(G, N)
        print("  %2d   %.4e   %.4e" % (N, e, taylor(G, N, off)), flush=True)
        with open("coef_%s_%d.txt" % (case, N), "w") as f:
            for a in c:
                f.write(format(a, ".40E") + "\n")   # Decimal.__format__, not float
```

</details>

### Number of terms

This table gives the number of terms for a truncation error of less than 0.3 eps. Above
this count, the arithmetic of the type gives more error than the polynomial. The Taylor
columns give the coefficients `2/(2k+1)`. The Remez columns give the coefficients from
the Remez program.

| type | option 2 Taylor | option 2 Remez | option 3 Taylor | option 3 Remez | terms in the table now |
| --- | --- | --- | --- | --- | --- |
| `decimal32_t` | 6 | 4 | 3 | 2 | 12 |
| `decimal64_t` | 15 | 10 | 9 | 6 | 20 |
| `decimal128_t` | 34 | 22 | 20 | 15 | 36 |

Option 2 and option 3 both need fewer coefficients than the tables contain now. Remez
coefficients decrease the count again. For the `decimal128_t` type they remove 12 terms
from option 2, and 5 terms from option 3.

The counts for the `decimal32_t` and `decimal64_t` types come from the Remez program.
The appendix gives a measurement for the `decimal128_t` type.

Note for option 3 and the `decimal32_t` and `decimal64_t` types: more terms do not give
less than 1 eps. The error of these two types stops at a lower limit near `x = 0.42`.
The section "Error and speed for each range" gives the cause.

## Benchmark results

The test compares the three methods with the current code. The test range is the full
branch, from `-0.99` to `0.5`. Each range in the tables below has 700 points, with more
points near zero.

The error columns are the same in each test. The time columns change by up to 15 percent
between tests. Use the times to compare the methods. Do not use them as exact values.

The test uses 7, 16 and 34 terms for option 2. It uses 4, 10 and 21 terms for option 3.
These counts are equal to or more than the counts in the table above.

### Summary of the full branch

| type | current | option 1 | option 2 | option 3 |
| `decimal32_t` error | 3.4e+05 eps | 2.39 eps | **0.88 eps** | 2.67 eps |
| `decimal32_t` speed | 548 ns | 634 ns | **451 ns** | 525 ns |
| `decimal64_t` error | 2.5e+14 eps | 3.03 eps | **0.96 eps** | 2.02 eps |
| `decimal64_t` speed | 3056 ns | 2642 ns | 2427 ns | **1938 ns** |
| `decimal128_t` error | 1.6e+32 eps | 4.22 eps | 0.77 eps | **0.53 eps** |
| `decimal128_t` speed | 5995 ns | 5380 ns | 5646 ns | **4762 ns** |

All three methods correct the problem. The current code has errors of 1e5 eps and more.
The three methods have errors of 5 eps and less.

Option 2 gives the smallest error for the `decimal32_t` and `decimal64_t` types. It is
also the fastest method for the `decimal32_t` type.

Option 3 is the fastest method for the `decimal64_t` and `decimal128_t` types. For the
`decimal128_t` type it also gives the smallest error. It is about 1.5 times faster than
the current code for the `decimal64_t` type.

### Error and speed for each range

The error columns give the largest error in the range. The speed columns give the
average time for one call.

**`decimal32_t`**

| range | current | option 1 | option 2 | option 3 | current ns | option 1 ns | option 2 ns | option 3 ns |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| \|x\| <= 0.1 | 0.864 | 0.864 | 0.792 | 0.485 | 540 | 571 | 363 | 391 |
| 0.1 < \|x\| <= 0.5 | 7.38 | 2.59 | 0.776 | 2.86 | 509 | 573 | 344 | 428 |
| -0.5 <= x < -0.1 | 11.9 | 1.35 | 0.794 | 1.19 | 505 | 573 | 349 | 520 |
| -0.99 < x < -0.5 | 3.37e+05 | 1.1 | 1.1 | 1.04 | 501 | 691 | 690 | 686 |
| full branch | 3.37e+05 | 2.39 | 0.879 | 2.67 | 548 | 634 | 451 | 525 |

**`decimal64_t`**

| range | current | option 1 | option 2 | option 3 | current ns | option 1 ns | option 2 ns | option 3 ns |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| \|x\| <= 0.1 | 0.951 | 0.951 | 0.906 | 0.466 | 3153 | 3218 | 2544 | 1753 |
| 0.1 < \|x\| <= 0.5 | 1.81e+07 | 3.58 | 0.915 | 2.8 | 2996 | 2416 | 2590 | 1954 |
| -0.5 <= x < -0.1 | 3e+07 | 3.5 | 0.985 | 2 | 3033 | 2468 | 2515 | 1926 |
| -0.99 < x < -0.5 | 2.52e+14 | 1.01 | 1.01 | 1.19 | 2956 | 2187 | 2211 | 2215 |
| full branch | 2.52e+14 | 3.03 | 0.964 | 2.02 | 3056 | 2642 | 2427 | 1938 |

**`decimal128_t`**

| range | current | option 1 | option 2 | option 3 | current ns | option 1 ns | option 2 ns | option 3 ns |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| \|x\| <= 0.1 | 0.93 | 0.93 | 0.735 | 0.497 | 5760 | 5970 | 6579 | 4579 |
| 0.1 < \|x\| <= 0.5 | 1.59e+20 | 2.36 | 0.715 | 0.448 | 4803 | 4400 | 4582 | 3651 |
| -0.5 <= x < -0.1 | 2.69e+20 | 4.81 | 0.717 | 0.45 | 4744 | 4470 | 4876 | 4292 |
| -0.99 < x < -0.5 | 1.61e+32 | 0.851 | 0.851 | 0.608 | 4750 | 4081 | 4082 | 5142 |
| full branch | 1.61e+32 | 4.22 | 0.766 | 0.526 | 5995 | 5380 | 5646 | 4762 |

In the range `-0.99 < x < -0.5`, option 1 and option 2 give the same values.
Both methods call `log(x + 1)` there. The errors of 1.1, 1.0 and 0.85 eps in this
range are the errors of the `log` function.

Option 1 gives errors of 1.4 to 4.8 eps in the middle ranges. These are also the errors
of the `log` function. Option 2 and option 3 keep a polynomial in these ranges.

For the `decimal32_t` and `decimal64_t` types, the error of option 3 stops at a lower
limit near `x = 0.42`. The largest errors are at `x = 0.4254` and `x = 0.4385`. These
points are only a small distance outside the range of step 1, so `k` becomes 1. The
result is then the sum of `ln2` and a negative value of almost the same size. This
calculation loses about 0.3 digits.

The glibc function has the same lower limit. Option 2 does not have it, because option 2
does not change the argument.

### Why log(1 + x) is not the correction for all values

Do not use `log(1 + x)` for all values in the branch. For small values of `|x|`,
this calculation loses digits. This is the reason for the `log1p` function. This table
shows the error of `log(1 + x)`. The unit is eps:

| x | -0.05 | 0.05 | 0.1 | 0.2 | 0.4 | 0.5 |
| --- | --- | --- | --- | --- | --- | --- |
| `decimal32_t` | 9.27 | 0.09 | 0.32 | 0.86 | 0.41 | 0.23 |
| `decimal64_t` | 10.40 | 3.34 | 0.84 | 0.14 | 0.09 | 0.20 |
| `decimal128_t` | 5.04 | 1.05 | 1.34 | 0.37 | 1.22 | 0.09 |

The error is small in the middle of the range. But the error becomes large for small
values of `|x|`. The code must use a polynomial near zero.

### Which option to use

Option 2 is the smallest change. It needs no new code to change the argument, and it
does not divide `ln2` into two parts. It gives an error of less than 1 eps for all three
types. It also makes the tables smaller.

Option 3 is better for the `decimal128_t` type. It is faster, and the error is smaller.
It also needs the smallest tables.

Use option 2 for the `decimal32_t` and `decimal64_t` types. Use option 3 for the
`decimal128_t` type. But option 2 without option 3 corrects the problem for all six
types.

## Appendix

<details>
<summary>Check of the correct values</summary>

To check the correct values, compare `boost::math::log1p<cpp_bin_float_50>` with
`mpfr_log1p`. The difference between the two libraries is less than 3.8e-49 at all 20
test points. The largest difference is at `x = -0.999`. This table shows some of the
points:

| x | -0.999 | -0.99 | -0.9 | -0.5 | -0.4 | -0.25 | -0.1 | 0.1 | 0.25 | 0.4 | 0.5 |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| difference | 3.8e-49 | 1.6e-49 | 1.7e-50 | 1.1e-51 | 1.1e-50 | 8.5e-51 | 2.4e-52 | 1.2e-51 | 1.8e-51 | 8.7e-51 | 6.1e-51 |

These differences are less than 0.55 units in the last place of the `cpp_bin_float_50`
type.
They are also 1e16 times smaller than 1 eps of the `decimal128_t` type.

Note on the measurement: the test moves the `cpp_bin_float_50` value into MPFR as a
decimal string. Print 55 digits or more. With fewer digits, the measurement gives the
error of the string, and not the difference between the two libraries.

The two libraries give the same values. This shows that the values from Boost.Decimal
are not correct.

</details>

<details>
<summary>Why the test does not find the problem</summary>

The test in `test/test_log1p.cpp` does not find the problem for three reasons:

- The tests of the `decimal64_t` and `decimal128_t` types use only values of `|x|` less
  than or equal to 0.1. The `test_log1p_64` and `test_log1p_128` functions use
  `x = n/100`, for `n` from `-10` to `10`. These values are inside the correct range
  of the two tables, which is 0.22 and 0.14. The tests do not use the incorrect part of
  the branch.
- The tests with random values use only the `decimal32_t` type. The type is fixed in
  the `test_log1p` function. This type has the smallest error, because it is correct to
  about `|x| = 0.42`. The range `[-0.375, 0.375]` is inside this correct range.

  The range `[0, 2]` goes into the incorrect part only between 0.426 and 0.5. The
  largest error there is 7.4 eps, which is 7.4e-06. The tolerance of this test is 96
  times the epsilon of the `float` type, which is 1.1e-05. The test does not fail, but
  the tolerance is only 1.5 times the error. The third range, `[1, 1e6]`, uses the `log`
  branch only.
- The tests with random values compare with the `float` type. They compare
  `std::log1p(float)` with `static_cast<float>(val_dec)`. The tolerance is
  `std::numeric_limits<float>::epsilon() * tol_factor`. A `float` value has about 7
  correct digits, which is not enough. It does not show an error of 1.8e-8 in the
  `decimal64_t` type. It also does not show an error of 1.6e-13 in the `decimal128_t`
  type.

The `test_log1p_128` function also uses `tol_factor = 8192`. This is a tolerance of
about 8e-30. With a large tolerance, the test does not find many errors.

A new test must follow these three rules:

1. Use each of the six types in the full branch, from `-1` to `0.5`.
2. Compare with correct values that have the precision of that type or more.
3. Use decimal strings for the correct values. The `test_log1p_128` function already
   does this. Do not use `double` or `float` values.

</details>

<details>
<summary>Measured results for the `decimal128_t` type</summary>

The test measures the largest error with `mpfr_log1p`, with the same method as the
benchmark above. The unit is eps = `1e-33`. The times come from one test on the same
machine.

Option 3, in the full branch from `-0.99` to `0.5`:

| coefficients | terms | largest error | at x | time |
| current code | 36 | 1.6e+32 eps | -0.99 | 5627 ns |
| Taylor | 19 | 5.97 eps | 0.414 | 4356 ns |
| Taylor | 20 | 0.632 eps | -0.789 | 4690 ns |
| Remez | 14 | 1.30 eps | -0.100 | 3658 ns |
| Remez | 15 | 0.632 eps | -0.789 | 3825 ns |

Option 2, in the range of its polynomial, from `-0.5` to `0.5`:

| coefficients | terms | largest error | at x | time |
| Taylor | 32 | 4.72 eps | -0.5 | 5604 ns |
| Taylor | 33 | 0.729 eps | 0.110 | 5761 ns |
| Remez | 21 | 1.15 eps | 0.002 | 3723 ns |
| Remez | 22 | 0.729 eps | 0.110 | 3789 ns |

Do not compare the two tables. They use different ranges.

With 15 Remez terms, option 3 gives the same error at the same point as with 20 or 21
Taylor terms. This error of 0.632 eps is the error of the decimal arithmetic. The
polynomial does not add to it. The same is true for option 2 with 22 Remez terms.

Compare the 113-bit function of glibc, in `sysdeps/ieee754/ldbl-128/s_log1pl.c`. It uses
a rational function `P(x)/Q(x)` and one division. `P` has 13 coefficients and `Q` has 13
coefficients. The highest coefficient of `Q` is 1, so the file lists 25 values. The file
gives a largest relative error of 1.9e-34 for the range `-1` to `8`.

Option 3 for the `decimal128_t` type gives 6.3e-34 with 20 Taylor coefficients, and the
same 6.3e-34 with 15 Remez coefficients. It needs no division.

</details>

<details>
<summary>The comment in the table is not correct</summary>

The `d128_coeffs` and `d128_fast_coeffs` tables have this comment:

```cpp
// Series[Log[(1 + (z/2))/(1 - (z/2))], {z, 0, 43}]
```

But the values in the tables are the Taylor series of `Log[1 + x]`. The first three
values are `-1/2`, `+1/3`, and `-1/4`. These are the same values as in the `d32` and
`d64` tables. The last value is `+1/37`.

The comment on each line of the file gives the powers `z^2` to `z^37`, which is 36
values. The series in the comment has odd powers of `z` only. It gives 21 terms up to
`z^43`.

The comment shows the formula `2*atanh(z/2)`. This formula is the correction in the
section "Option 2: change the variable". The comment is correct for that method. But
the coefficients are the coefficients of `Log[1 + x]`.

</details>

ibmibmibm:
The log1p function calculated a Taylor polynomial of log(1 + x) for all values of x in the range (-1, 0.5]. The test had an upper limit at x = 0.5, but it had no lower limit. The coefficient tables were not long enough for this range.

The function gave a result with too few correct digits, and it reported no error. At x = -0.99, the relative error was 34 percent for decimal32_t. It was 25 percent for decimal64_t, and 16 percent for decimal128_t. All six types had this problem, because the fast types hold a copy of the same values.

The tables now hold the coefficients 2/(2k+1) of the series 2 * atanh(w), with w = x / (2 + x). The counts are 7, 16 and 34 terms, in place of 12, 20 and 36. For |x| not more than 1/2, the value of |w| is not more than 1/3. A test of abs(x) gives the lower limit that the code did not have. For all larger values of |x|, the code calculates log(x + 1).

For |x| less than 10^-digits10, the function now returns x. The sum of the terms after x is less than one half of the last digit of x. This branch also keeps w * w away from the exponents where fma gives an incorrect result for decimal32_t and decimal64_t. The exponents are 1e-21 to 1e-31 for decimal32_t, and 1e-42 to 1e-62 for decimal64_t.

A test with MPFR at 1610 points of the branch gives these results:

  decimal32_t    3.37e+05 eps -> 1.02 eps,   551 ns -> 478 ns
  decimal64_t    2.52e+14 eps -> 0.88 eps,  3299 ns -> 2613 ns
  decimal128_t   1.61e+32 eps -> 0.91 eps,  5291 ns -> 5291 ns

The largest error is now at x = -0.64. This point is in the branch that calculates log(x + 1), thus the value is the error of the log function.

The atanh, asinh and acosh functions call log1p. They get the correction with no change. For decimal128_t, the error of atanh at x = 0.49 decreases from 5.13e+19 eps to 0.23 eps. The error of acosh at x = 1.08 decreases from 6.45e+19 eps to 0.41 eps.

The old tests for these four functions used decimal32_t and compared with float. A test of this form cannot find an error that is smaller than the epsilon of float. The old tests also did not use values of x that are less than -0.375. Each test now has a table of control values with 36 digits, for all six types. The table for log1p has 32 points, and four of them are near zero. Each new test fails with the old code.

Fixes #1430.

### DIFF
--- a/include/boost/decimal/detail/cmath/impl/log1p_impl.hpp
+++ b/include/boost/decimal/detail/cmath/impl/log1p_impl.hpp
@@ -26,185 +26,157 @@ template <bool b>
 struct log1p_table_imp
 {
 private:
-    using d32_coeffs_t  = std::array<decimal32_t,  12>;
-    using d64_coeffs_t  = std::array<decimal64_t,  20>;
-    using d128_coeffs_t = std::array<decimal128_t, 36>;
+    using d32_coeffs_t  = std::array<decimal32_t,   7>;
+    using d64_coeffs_t  = std::array<decimal64_t,  16>;
+    using d128_coeffs_t = std::array<decimal128_t, 34>;
 
-    using d32_fast_coeffs_t  = std::array<decimal_fast32_t,  12>;
-    using d64_fast_coeffs_t  = std::array<decimal_fast64_t,  20>;
-    using d128_fast_coeffs_t = std::array<decimal_fast128_t, 36>;
+    using d32_fast_coeffs_t  = std::array<decimal_fast32_t,   7>;
+    using d64_fast_coeffs_t  = std::array<decimal_fast64_t,  16>;
+    using d128_fast_coeffs_t = std::array<decimal_fast128_t, 34>;
 
 public:
     static constexpr d32_coeffs_t d32_coeffs =
     {{
-         // Series[Log[1 + x], {x, 0, 13}]
-         //            (1),                                                     // * z
-         -boost::decimal::decimal32_t { 5, -1 },                                  // * z^2
-          boost::decimal::decimal32_t { UINT64_C(3333333333333333333), -19 },     // * z^3
-         -boost::decimal::decimal32_t { 25, -2 },                                 // * z^4
-          boost::decimal::decimal32_t { 2, -1 },                                  // * z^5
-         -boost::decimal::decimal32_t { UINT64_C(1666666666666666667), -19 },     // * z^6
-          boost::decimal::decimal32_t { UINT64_C(1428571428571428571), -19 },     // * z^7
-         -boost::decimal::decimal32_t { 125, -3 },                                // * z^8
-          boost::decimal::decimal32_t { UINT64_C(1111111111111111111), -19 },     // * z^9
-         -boost::decimal::decimal32_t { 1, -1 },                                  // * z^10
-          boost::decimal::decimal32_t { UINT64_C(9090909090909090909), -19 - 1 }, // * z^11
-         -boost::decimal::decimal32_t { UINT64_C(8333333333333333333), -19 - 1 }, // * z^12
-          boost::decimal::decimal32_t { UINT64_C(7692307692307692308), -19 - 1 }, // * z^13
+         // Series[2*ArcTanh[w], {w, 0, 13}]
+         boost::decimal::decimal32_t { 2, 0 },                               // * w^1
+         boost::decimal::decimal32_t { UINT64_C(6666666666666666667), -19 }, // * w^3
+         boost::decimal::decimal32_t { UINT64_C(4000000000000000000), -19 }, // * w^5
+         boost::decimal::decimal32_t { UINT64_C(2857142857142857143), -19 }, // * w^7
+         boost::decimal::decimal32_t { UINT64_C(2222222222222222222), -19 }, // * w^9
+         boost::decimal::decimal32_t { UINT64_C(1818181818181818182), -19 }, // * w^11
+         boost::decimal::decimal32_t { UINT64_C(1538461538461538462), -19 }, // * w^13
     }};
 
     static constexpr d32_fast_coeffs_t d32_fast_coeffs =
     {{
-         // Series[Log[1 + x], {x, 0, 13}]
-         //            (1),                                                     // * z
-         -boost::decimal::decimal_fast32_t { 5, -1 },                                  // * z^2
-          boost::decimal::decimal_fast32_t { UINT64_C(3333333333333333333), -19 },     // * z^3
-         -boost::decimal::decimal_fast32_t { 25, -2 },                                 // * z^4
-          boost::decimal::decimal_fast32_t { 2, -1 },                                  // * z^5
-         -boost::decimal::decimal_fast32_t { UINT64_C(1666666666666666667), -19 },     // * z^6
-          boost::decimal::decimal_fast32_t { UINT64_C(1428571428571428571), -19 },     // * z^7
-         -boost::decimal::decimal_fast32_t { 125, -3 },                                // * z^8
-          boost::decimal::decimal_fast32_t { UINT64_C(1111111111111111111), -19 },     // * z^9
-         -boost::decimal::decimal_fast32_t { 1, -1 },                                  // * z^10
-          boost::decimal::decimal_fast32_t { UINT64_C(9090909090909090909), -19 - 1 }, // * z^11
-         -boost::decimal::decimal_fast32_t { UINT64_C(8333333333333333333), -19 - 1 }, // * z^12
-          boost::decimal::decimal_fast32_t { UINT64_C(7692307692307692308), -19 - 1 }, // * z^13
-     }};
+         // Series[2*ArcTanh[w], {w, 0, 13}]
+         boost::decimal::decimal_fast32_t { 2, 0 },                               // * w^1
+         boost::decimal::decimal_fast32_t { UINT64_C(6666666666666666667), -19 }, // * w^3
+         boost::decimal::decimal_fast32_t { UINT64_C(4000000000000000000), -19 }, // * w^5
+         boost::decimal::decimal_fast32_t { UINT64_C(2857142857142857143), -19 }, // * w^7
+         boost::decimal::decimal_fast32_t { UINT64_C(2222222222222222222), -19 }, // * w^9
+         boost::decimal::decimal_fast32_t { UINT64_C(1818181818181818182), -19 }, // * w^11
+         boost::decimal::decimal_fast32_t { UINT64_C(1538461538461538462), -19 }, // * w^13
+    }};
 
     static constexpr d64_coeffs_t d64_coeffs =
     {{
-         // Series[Log[1 + x], {x, 0, 21}]
-         //            (1),                                                     // * z
-         -boost::decimal::decimal64_t { 5, -1 },                                  // * z^2
-          boost::decimal::decimal64_t { UINT64_C(3333333333333333333), -19 },     // * z^3
-         -boost::decimal::decimal64_t { 25, -2 },                                 // * z^4
-          boost::decimal::decimal64_t { 2, -1 },                                  // * z^5
-         -boost::decimal::decimal64_t { UINT64_C(1666666666666666667), -19 },     // * z^6
-          boost::decimal::decimal64_t { UINT64_C(1428571428571428571), -19 },     // * z^7
-         -boost::decimal::decimal64_t { 125, -3 },                                // * z^8
-          boost::decimal::decimal64_t { UINT64_C(1111111111111111111), -19 },     // * z^9
-         -boost::decimal::decimal64_t { 1, -1 },                                  // * z^10
-          boost::decimal::decimal64_t { UINT64_C(9090909090909090909), -19 - 1 }, // * z^11
-         -boost::decimal::decimal64_t { UINT64_C(8333333333333333333), -19 - 1 }, // * z^12
-          boost::decimal::decimal64_t { UINT64_C(7692307692307692308), -19 - 1 }, // * z^13
-         -boost::decimal::decimal64_t { UINT64_C(7142857142857142857), -19 - 1 }, // * z^14
-          boost::decimal::decimal64_t { UINT64_C(6666666666666666667), -19 - 1 }, // * z^15
-         -boost::decimal::decimal64_t { UINT64_C(6250000000000000000), -19 - 1 }, // * z^16
-          boost::decimal::decimal64_t { UINT64_C(5882352941176470588), -19 - 1 }, // * z^17
-         -boost::decimal::decimal64_t { UINT64_C(5555555555555555556), -19 - 1 }, // * z^18
-          boost::decimal::decimal64_t { UINT64_C(5263157894736842105), -19 - 1 }, // * z^19
-         -boost::decimal::decimal64_t { 5, -2 },                                  // * z^20
-          boost::decimal::decimal64_t { UINT64_C(4761904761904761905), -19 - 1 }, // * z^21
-     }};
+         // Series[2*ArcTanh[w], {w, 0, 31}]
+         boost::decimal::decimal64_t { 2, 0 },                                   // * w^1
+         boost::decimal::decimal64_t { UINT64_C(6666666666666666667), -19 },     // * w^3
+         boost::decimal::decimal64_t { UINT64_C(4000000000000000000), -19 },     // * w^5
+         boost::decimal::decimal64_t { UINT64_C(2857142857142857143), -19 },     // * w^7
+         boost::decimal::decimal64_t { UINT64_C(2222222222222222222), -19 },     // * w^9
+         boost::decimal::decimal64_t { UINT64_C(1818181818181818182), -19 },     // * w^11
+         boost::decimal::decimal64_t { UINT64_C(1538461538461538462), -19 },     // * w^13
+         boost::decimal::decimal64_t { UINT64_C(1333333333333333333), -19 },     // * w^15
+         boost::decimal::decimal64_t { UINT64_C(1176470588235294118), -19 },     // * w^17
+         boost::decimal::decimal64_t { UINT64_C(1052631578947368421), -19 },     // * w^19
+         boost::decimal::decimal64_t { UINT64_C(9523809523809523810), -19 - 1 }, // * w^21
+         boost::decimal::decimal64_t { UINT64_C(8695652173913043478), -19 - 1 }, // * w^23
+         boost::decimal::decimal64_t { UINT64_C(8000000000000000000), -19 - 1 }, // * w^25
+         boost::decimal::decimal64_t { UINT64_C(7407407407407407407), -19 - 1 }, // * w^27
+         boost::decimal::decimal64_t { UINT64_C(6896551724137931034), -19 - 1 }, // * w^29
+         boost::decimal::decimal64_t { UINT64_C(6451612903225806452), -19 - 1 }, // * w^31
+    }};
 
     static constexpr d64_fast_coeffs_t d64_fast_coeffs =
     {{
-         // Series[Log[1 + x], {x, 0, 21}]
-         //            (1),                                                     // * z
-         -boost::decimal::decimal_fast64_t { 5, -1 },                                  // * z^2
-          boost::decimal::decimal_fast64_t { UINT64_C(3333333333333333333), -19 },     // * z^3
-         -boost::decimal::decimal_fast64_t { 25, -2 },                                 // * z^4
-          boost::decimal::decimal_fast64_t { 2, -1 },                                  // * z^5
-         -boost::decimal::decimal_fast64_t { UINT64_C(1666666666666666667), -19 },     // * z^6
-          boost::decimal::decimal_fast64_t { UINT64_C(1428571428571428571), -19 },     // * z^7
-         -boost::decimal::decimal_fast64_t { 125, -3 },                                // * z^8
-          boost::decimal::decimal_fast64_t { UINT64_C(1111111111111111111), -19 },     // * z^9
-         -boost::decimal::decimal_fast64_t { 1, -1 },                                  // * z^10
-          boost::decimal::decimal_fast64_t { UINT64_C(9090909090909090909), -19 - 1 }, // * z^11
-         -boost::decimal::decimal_fast64_t { UINT64_C(8333333333333333333), -19 - 1 }, // * z^12
-          boost::decimal::decimal_fast64_t { UINT64_C(7692307692307692308), -19 - 1 }, // * z^13
-         -boost::decimal::decimal_fast64_t { UINT64_C(7142857142857142857), -19 - 1 }, // * z^14
-          boost::decimal::decimal_fast64_t { UINT64_C(6666666666666666667), -19 - 1 }, // * z^15
-         -boost::decimal::decimal_fast64_t { UINT64_C(6250000000000000000), -19 - 1 }, // * z^16
-          boost::decimal::decimal_fast64_t { UINT64_C(5882352941176470588), -19 - 1 }, // * z^17
-         -boost::decimal::decimal_fast64_t { UINT64_C(5555555555555555556), -19 - 1 }, // * z^18
-          boost::decimal::decimal_fast64_t { UINT64_C(5263157894736842105), -19 - 1 }, // * z^19
-         -boost::decimal::decimal_fast64_t { 5, -2 },                                  // * z^20
-          boost::decimal::decimal_fast64_t { UINT64_C(4761904761904761905), -19 - 1 }, // * z^21
-     }};
+         // Series[2*ArcTanh[w], {w, 0, 31}]
+         boost::decimal::decimal_fast64_t { 2, 0 },                                   // * w^1
+         boost::decimal::decimal_fast64_t { UINT64_C(6666666666666666667), -19 },     // * w^3
+         boost::decimal::decimal_fast64_t { UINT64_C(4000000000000000000), -19 },     // * w^5
+         boost::decimal::decimal_fast64_t { UINT64_C(2857142857142857143), -19 },     // * w^7
+         boost::decimal::decimal_fast64_t { UINT64_C(2222222222222222222), -19 },     // * w^9
+         boost::decimal::decimal_fast64_t { UINT64_C(1818181818181818182), -19 },     // * w^11
+         boost::decimal::decimal_fast64_t { UINT64_C(1538461538461538462), -19 },     // * w^13
+         boost::decimal::decimal_fast64_t { UINT64_C(1333333333333333333), -19 },     // * w^15
+         boost::decimal::decimal_fast64_t { UINT64_C(1176470588235294118), -19 },     // * w^17
+         boost::decimal::decimal_fast64_t { UINT64_C(1052631578947368421), -19 },     // * w^19
+         boost::decimal::decimal_fast64_t { UINT64_C(9523809523809523810), -19 - 1 }, // * w^21
+         boost::decimal::decimal_fast64_t { UINT64_C(8695652173913043478), -19 - 1 }, // * w^23
+         boost::decimal::decimal_fast64_t { UINT64_C(8000000000000000000), -19 - 1 }, // * w^25
+         boost::decimal::decimal_fast64_t { UINT64_C(7407407407407407407), -19 - 1 }, // * w^27
+         boost::decimal::decimal_fast64_t { UINT64_C(6896551724137931034), -19 - 1 }, // * w^29
+         boost::decimal::decimal_fast64_t { UINT64_C(6451612903225806452), -19 - 1 }, // * w^31
+    }};
 
     static constexpr d128_coeffs_t d128_coeffs =
     {{
-         // Series[Log[(1 + (z/2))/(1 - (z/2))], {z, 0, 43}]
-         //            (1),                                                                                                                    // * z
-         -::boost::decimal::decimal128_t { 5, -1 },                                                                                              // * z^2
-          ::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(180700362080917), UINT64_C(7483252092553221458)  }, -34 }, // * z^3
-         -::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(135525271560688), UINT64_C(1000753050987528192)  }, -34 }, // * z^4
-          ::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(108420217248550), UINT64_C(8179300070273843200)  }, -34 }, // * z^5
-         -::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(90350181040458),  UINT64_C(12964998083131386532) }, -34 }, // * z^6
-          ::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(77443012320393),  UINT64_C(3207108039665666332)  }, -34 }, // * z^7
-         -::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(67762635780344),  UINT64_C(500376525493764096)   }, -34 }, // * z^8
-          ::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(60233454026972),  UINT64_C(8643332055420924359)  }, -34 }, // * z^9
-         -::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(54210108624275),  UINT64_C(4089650035136921600)  }, -34 }, // * z^10
-          ::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(492819169311592), UINT64_C(17054915875379776418) }, -35 }, // * z^11
-         -::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(451750905202293), UINT64_C(9484758194528277842)  }, -35 }, // * z^12
-          ::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(417000835571347), UINT64_C(15850062977145160938) }, -35 }, // * z^13
-         -::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(387215061601965), UINT64_C(16035540198328331700) }, -35 }, // * z^14
-          ::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(361400724161834), UINT64_C(14966504185106442916) }, -35 }, // * z^15
-         -::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(338813178901720), UINT64_C(2501882627468820480)  }, -35 }, // * z^16
-          ::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(318882991907501), UINT64_C(5610020838860575434)  }, -35 }, // * z^17
-         -::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(301167270134862), UINT64_C(6323172129685518558)  }, -35 }, // * z^18
-          ::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(285316361180395), UINT64_C(16670067533954968520) }, -35 }, // * z^19
-         -::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(271050543121376), UINT64_C(2001506101975056384)  }, -35 }, // * z^20
-          ::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(258143374401310), UINT64_C(10690360132218887800) }, -35 }, // * z^21
-         -::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(246409584655796), UINT64_C(8527457937689888204)  }, -35 }, // * z^22
-          ::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(235696124453370), UINT64_C(9760763598982462770)  }, -35 }, // * z^23
-         -::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(225875452601146), UINT64_C(13965751134118914724) }, -35 }, // * z^24
-          ::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(216840434497100), UINT64_C(16358600140547686400) }, -35 }, // * z^25
-         -::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(208500417785673), UINT64_C(17148403525427356272) }, -35 }, // * z^26
-          ::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(200778180089908), UINT64_C(4215448086457012372)  }, -35 }, // * z^27
-         -::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(193607530800982), UINT64_C(17241142136018941658) }, -35 }, // * z^28
-          ::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(186931409049224), UINT64_C(16646619993397598836) }, -35 }, // * z^29
-         -::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(180700362080917), UINT64_C(7483252092553221458)  }, -35 }, // * z^30
-          ::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(174871318142823), UINT64_C(5456688082434451252)  }, -35 }, // * z^31
-         -::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(169406589450860), UINT64_C(1250941313734410240)  }, -35 }, // * z^32
-          ::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(164273056437197), UINT64_C(11833886649696442678) }, -35 }, // * z^33
-         -::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(159441495953750), UINT64_C(12028382456285063520) }, -35 }, // * z^34
-          ::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(154886024640786), UINT64_C(6414216079331332674)  }, -35 }, // * z^35
-         -::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(150583635067431), UINT64_C(3161586064842759274)  }, -35 }, // * z^36
-          ::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(146513807092635), UINT64_C(13545911456276754540) }, -35 }, // * z^37
+         // Series[2*ArcTanh[w], {w, 0, 67}]
+         ::boost::decimal::decimal128_t { 2, 0 },                                                                                        // * w^1
+         ::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(361400724161834), UINT64_C(14966504185106442923) }, -34 }, // * w^3
+         ::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(216840434497100), UINT64_C(16358600140547686400) }, -34 }, // * w^5
+         ::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(154886024640786), UINT64_C(6414216079331332681) }, -34 },  // * w^7
+         ::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(120466908053944), UINT64_C(17286664110841848718) }, -34 }, // * w^9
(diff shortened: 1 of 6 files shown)


mborland (approved):
This seems good to me unless you have any objections @ckormanyos

PR #1443, opened 2026-09-02
Make log1p increase at every argument

### LINKED ISSUE #1442: The log1p function does not always increase
## Summary

The mathematical function `log1p(z)` increases at every value of `z`. The
`boost::decimal::log1p` implementation decreases at some pairs of adjacent values of
`z`. These are two adjacent values of the `decimal32_t` type:

```
z            = 3.000025e-01   log1p = 2.623664e-01   correct = 2.623662e-01
nextafter(z) = 3.000026e-01   log1p = 2.623662e-01   correct = 2.623663e-01
```

The argument increases by one unit of the last place (ulp), and the result decreases by
two ulp. A pair of this kind is an inversion.

The error of each of the two results is one ulp. Thus this is not a defect of accuracy.
An error of one ulp is enough to make an inversion, because the slope of `log1p` is less
than 1 in this range. Two adjacent arguments can have correct results which differ by
less than one ulp. An error of one ulp can then put them in the wrong order.

A walk of all 58000000 representable values of the `decimal32_t` type from 1e-7 to 0.5
gives 125402 inversions. This report calls that walk the full walk of the positive side.
The full walk of the negative side, from -0.5 to -1e-7, gives 0. Only a positive
argument has this defect.

All six decimal types have inversions. This table counts them at 40000 consecutive
values from each of the six points 0.05, 0.1, 0.2, 0.3, 0.4 and 0.45:

| type | inversions of 240000 pairs |
| --- | --- |
| `decimal32_t` | 1242 |
| `decimal_fast32_t` | 1242 |
| `decimal64_t` | 1215 |
| `decimal_fast64_t` | 1215 |
| `decimal128_t` | 1250 |
| `decimal_fast128_t` | 1250 |

Code which depends on the order of the results can fail. A bisection, a search of a
table and a test of an inverse function each need that order.

The cause is the argument reduction `w = x / (2 + x)`. The value of `x` is in the
numerator and in the denominator. When `x` increases, the numerator makes `w` larger and
the rounded denominator makes `w` smaller. The two effects are in opposition, and the
quotient can decrease.

This report gives three corrections:
- Correction A changes the form of the reduction.
- Correction B corrects the rounding of the reduction.
- Correction C corrects the rounding of the sum of the result.

Correction A is the form `1 / (1 + 2/x)`, which is equal to `x / (2 + x)` and has `x` in
one place only. It needs one line, and it removes every inversion of all six types, at a
cost of not more than 3 percent. It does not make the error smaller.

Correction B keeps the sum and corrects it. The error of the sum `2 + x` is exact, and
the residual of the quotient is exact, thus `w` gets the correct rounding of
`x / (2 + x)`. A value which has the correct rounding cannot decrease, because `log1p`
increases. Correction B alone removes every inversion of all six types.

Correction C gets the result right. The first coefficient of the series is exactly 2,
thus the result is `2*w` plus a small tail. The reduction stays in two parts as
`wh + wl`, and the result is `2*wh + (2*wl + tail)`. Only the last sum then rounds at
the size of the result. Correction C alone does not correct the order. It needs a
reduction which keeps the order of `x`, and correction A and correction B each give one.

This report writes the two pairs as B and C, and A and C. The unit of the error is eps,
which is `10^(1-digits10)` of relative error. With B and C the mean error of the sample
goes to 0.000 eps for the 32 and 64 bit types. For `decimal128_t` the pair needs a
product from a split to get there. Of the sample for `decimal32_t`, 99.9 percent then
get the correct rounding, against 48.2 percent for correction B alone. The pair A and C
gives 59.2 percent, and it keeps the small cost of correction A.

The cost of the pair B and C is 66 percent for `decimal32_t` and 18 percent for
`decimal64_t`. For `decimal128_t` it is 13 percent with the library `fma`, and 60
percent with a product from a split. The first leaves 803 inversions, and the second
leaves 0. The pair A and C costs 40 percent, 31 percent and 7 percent.

A polynomial of Remez with fewer terms pays for the correction. With it, the pair B and
C is faster than the current code by 5 percent for `decimal64_t`. For `decimal128_t` it
costs 30 percent, because that type also needs a product from a split.

A paper of 1990 gives the same two corrections for the same reduction, for binary
arithmetic. The section on published work names that paper.

Correction C needs a multiplication and an addition with one rounding. The `decimal32_t`
and `decimal64_t` types have one. The 128 bit types do not, because `d128_fma_impl`
calculates `x * y + z` with two roundings. With `d128_fma_impl` the pair B and C gives
232 inversions of `asinh` for the 128 bit types. With a product from the split of
Veltkamp it gives 0.

Thus `d128_fma_impl` needs a correction first. The section on the fused multiply and add
measures that defect.

Five other corrections were tested, and the appendix gives them. Correction D and
correction E need a type which is larger than the result, thus neither applies to
`decimal128_t`. The other three have a worse order, or a larger error, or a larger cost.
A polynomial of Remez, which is the best polynomial of its degree, gives the same error
as the polynomial of the current code. Thus the series is not a part of this defect. A
quotient of Pade puts the inversions back, because it has the same opposition of a
numerator and a denominator.

These are the choices:
- Correction A is one line. It removes every inversion of all six types at a cost of not
  more than 3 percent, and it does not make the error smaller.
- The pair B and C removes every inversion and gives a mean error of 0.000 eps. It costs
  66 percent for `decimal32_t` and 18 percent for `decimal64_t`.
- The pair B and C on the 128 bit types needs a correction of `d128_fma_impl` first.

The `asinh` function calls `log1p`, and it takes the inversions from it.

## Environment

- Boost.Decimal at commit `1ef35805` of `develop`
- MPFR 4.2.2 for the correct values, with `mpfr_log1p` at 300 bits.

## The measurements

The `log1p` function has two branches. For `|x| < 1/2` it uses an argument reduction and
a series. For every larger `|x|` it calculates `log(x + 1)`. This report examines the
branch of the reduction.

The tables of this report use these samples:
- The count of the inversions comes from 18 starting points, with 20000 consecutive
  values at each point.
- The full walk covers every representable value of the branch of the `decimal32_t`
  type. Each side of the branch has 58000000 values.
- The error comes from 2000 values of `x` with `0.000001 <= |x| <= 0.5`.
- The count of the correct roundings of `w` comes from 20000 values of the same shape.
- The time is the best of 15 runs of 50 passes.

The unit of the error is eps, which is `10^(1-digits10)` of relative error. One unit of
the last place of the result is `1/m` eps, where `m` is the first digit of the result. A
result which starts with 1 gives 1.00 eps for an error of one unit of the last place. A
result which starts with 9 gives 0.11 eps for the same error.

## Test program

This program needs Boost.Decimal and MPFR. It does not contain the correct values. It
calculates them with MPFR when it runs. Build it with
`g++ -std=c++17 -O2 log1p_mono_test.cpp -o log1p_mono_test -lmpfr`.

The program walks the consecutive values of the type from each starting point, and it
counts the pairs for which the result decreases. The walk covers a different span of `z`
for each type, because the unit of the last place is different. Thus the counts of the
three types are not directly comparable.

<details>

```cpp
// log1p is an increasing function. This program counts the pairs of adjacent
// arguments for which the implementation decreases, and it prints the first
// such pair for each type with the correctly rounded values from MPFR.
#include <boost/decimal.hpp>
#include <mpfr.h>
#include <cstdio>
#include <cstring>
#include <limits>
#include <string>

template <typename T>
static std::string to_scientific_string(T v)
{
    char b[160];
    auto r = to_chars(b, b + 158, v, chars_format::scientific, std::numeric_limits<T>::digits10 - 1);
    *r.ptr = 0;
    return std::string(b);
}

template <typename T>
static T from_scientific_string(const char* p)
{
    T o {};
    from_chars(p, p + std::strlen(p), o);
    return o;
}

template <typename T>
static T next_up(T x)
{
    return nextafter(x, std::numeric_limits<T>::infinity());
}

static const char* const STARTS[] = {
    "1e-3", "3e-3", "5e-3", "7e-3", "9e-3", "1e-2", "3e-2", "5e-2", "7e-2", "9e-2", "1e-1", "3e-1", "5e-1",
};

template <typename T>
static void count_decreasing_pairs(const char* name)
{
    std::printf("\n%s\n", name);
    std::printf("  %-8s %s\n", "from z =", "pairs that decrease, of 40000");
    for (const char* st : STARTS)
    {
        T z = from_scientific_string<T>(st);
        T p = log1p(z);
        long inv = 0;
        for (int i = 0; i < 40000; ++i)
        {
            z = next_up(z);
            const T v = log1p(z);
            if (v < p)
            {
                ++inv;
            }
            p = v;
        }
        std::printf("  %-8s %ld\n", st, inv);
    }
}

template <typename T>
static void show_first_pair(const char* name)
{
    constexpr int dg = std::numeric_limits<T>::digits10;
    mpfr_t a, r;
    mpfr_inits2(300, a, r, (mpfr_ptr)0);

    T z = from_scientific_string<T>("0.3");
    T p = log1p(z);
    T pz = z;
    for (int i = 0; i < 40000; ++i)
    {
        z = next_up(z);
        const T v = log1p(z);
        if (v < p)
        {
            char b1[200], b2[200];
            mpfr_set_str(a, to_scientific_string(pz).c_str(), 10, MPFR_RNDN);
            mpfr_log1p(r, a, MPFR_RNDN);
            mpfr_sprintf(b1, "%.*Re", dg - 1, r);
            mpfr_set_str(a, to_scientific_string(z).c_str(), 10, MPFR_RNDN);
            mpfr_log1p(r, a, MPFR_RNDN);
            mpfr_sprintf(b2, "%.*Re", dg - 1, r);
            std::printf("\n%s\n", name);
            std::printf("  z            = %s\n    log1p   = %s\n    correct = %s\n",
                        to_scientific_string(pz).c_str(), to_scientific_string(p).c_str(), b1);
            std::printf("  nextafter(z) = %s\n    log1p   = %s\n    correct = %s\n",
                        to_scientific_string(z).c_str(), to_scientific_string(v).c_str(), b2);
            mpfr_clears(a, r, (mpfr_ptr)0);
            return;
        }
        p = v;
        pz = z;
    }
    std::printf("%s: no such pair\n", name);
    mpfr_clears(a, r, (mpfr_ptr)0);
}

int main()
{
    std::printf("== count of the pairs that decrease ==\n");
    count_decreasing_pairs<decimal32_t>("decimal32_t");
    count_decreasing_pairs<decimal64_t>("decimal64_t");
    count_decreasing_pairs<decimal128_t>("decimal128_t");
    std::printf("\n== the first such pair, from z = 0.3 ==\n");
    show_first_pair<decimal32_t>("decimal32_t");
    show_first_pair<decimal64_t>("decimal64_t");
    show_first_pair<decimal128_t>("decimal128_t");
    return 0;
}
```

</details>

### Output of the program

<details>
<summary>Show the output</summary>

```
== count of the pairs that decrease ==

decimal32_t
  from z = pairs that decrease, of 40000
  1e-3     0
  3e-3     1
  5e-3     40
  7e-3     40
  9e-3     40
  1e-2     0
  3e-2     90
  5e-2     288
  7e-2     400
  9e-2     400
  1e-1     0
  3e-1     553
  5e-1     0

decimal64_t
  from z = pairs that decrease, of 40000
  1e-3     0
  3e-3     0
  5e-3     17
  7e-3     40
  9e-3     40
  1e-2     0
  3e-2     98
  5e-2     285
  7e-2     400
  9e-2     400
  1e-1     0
  3e-1     530
  5e-1     1

decimal128_t
  from z = pairs that decrease, of 40000
  1e-3     0
  3e-3     40
  5e-3     12
  7e-3     40
  9e-3     40
  1e-2     0
  3e-2     93
  5e-2     279
  7e-2     400
  9e-2     400
  1e-1     0
  3e-1     531
  5e-1     1

== the first such pair, from z = 0.3 ==

decimal32_t
  z            = 3.000025e-01
    log1p   = 2.623664e-01
    correct = 2.623662e-01
  nextafter(z) = 3.000026e-01
    log1p   = 2.623662e-01
    correct = 2.623663e-01

decimal64_t
  z            = 3.000000000000034e-01
    log1p   = 2.623642644674938e-01
    correct = 2.623642644674937e-01
  nextafter(z) = 3.000000000000035e-01
    log1p   = 2.623642644674936e-01
    correct = 2.623642644674937e-01

decimal128_t
  z            = 3.000000000000000000000000000000065e-01
    log1p   = 2.623642644674910520354959868809596e-01
    correct = 2.623642644674910520354959868809594e-01
  nextafter(z) = 3.000000000000000000000000000000066e-01
    log1p   = 2.623642644674910520354959868809593e-01
    correct = 2.623642644674910520354959868809595e-01
```

</details>

The rows at 1e-3, 1e-2 and 1e-1 have no inversion, and the rows at 3e-3, 3e-2 and 3e-1
have many. The cause of this pattern is exact.

## Cause

This is the branch, in `include/boost/decimal/detail/cmath/log1p.hpp`:

```cpp
// log1p(x) = 2 * atanh(w), with w = x / (2 + x).
// For |x| not more than 1/2, the value of |w| is not more than 1/3.
const T w { x / (T { 2, 0 } + x) };

result = w * detail::log1p_series_expansion(w * w);
```

These are the two values of the pair of the summary, with the denominator:

```
x        = 3.000025e-01   2 + x = 2.300002e+00   x/(2+x) = 1.304358e-01
next(x)  = 3.000026e-01   2 + x = 2.300003e+00   x/(2+x) = 1.304357e-01
```

The value of `x` increases by 1e-7, which is 3.33e-07 in proportion. The value of
`2 + x` increases by 1e-6, which is 4.35e-07 in proportion. The denominator grows faster
than the numerator, and the quotient decreases.

An inversion is possible when this is true:

```
ulp(2 + x) / (2 + x)   >   ulp(x) / x
```

For `x` in `[0.1, 1)` the value of `ulp(2 + x)` is 10 times `ulp(x)`, and the rule gives
`10*x > 2 + x`, thus `x > 2/9`. For `x` in `[0.01, 0.1)` the factor is 100, and the rule
gives `x > 2/99`. For `x` in `[0.001, 0.01)` the factor is 1000, and the rule gives
`x > 2/999`. The measurement agrees with each of the three limits:

```
decimal32_t, x/(2+x), 40000 consecutive values from each point
  predicted boundary 2/999 = 0.002002    from 0.0019   :     0   from 0.0021   :     1
  predicted boundary 2/99 = 0.020202     from 0.019    :     0   from 0.021    :     8
  predicted boundary 2/9 = 0.222222      from 0.21     :     0   from 0.23     :    68
```
A negative argument has no inversion, and the same mechanism gives the reason. For a
negative value of `x` the numerator is negative. When `x` increases, the numerator
becomes smaller in magnitude and it makes `w` larger. The denominator `2 + x` also
increases, and it makes the magnitude of `w` smaller, thus it makes `w` larger too. The
two effects are in the same direction, and no rounding can make the quotient decrease.
For a positive value of `x` the two effects are in opposition.

The division is the first step of the branch, and it makes all of the inversions. This
table walks 40000 consecutive values of `x` from 0.3, and it counts the inversions of
each step:

| step | `decimal32_t` | `decimal64_t` | `decimal128_t` |
| --- | --- | --- | --- |
| `w = x/(2+x)` | 553 | 530 | 531 |
| `w*w` | 553 | 530 | 531 |
| `log1p_series_expansion(w*w)` | 11 | 15 | 11 |
| `w * log1p_series_expansion(w*w)` | 553 | 530 | 531 |
| `log1p(x)` | 553 | 530 | 531 |

The last row has the count of `w`. Thus the series and the multiplication make no new
inversion. The series row alone has fewer, because the value of the series is near 2 and
its rounding removes most of the changes of `w*w`. A correction of the division corrects
the function.

A measurement separates the two roundings of the division. A form which corrects the
rounding of the quotient `x / d`, and which keeps the rounded sum `d`, gives 125402
inversions at the full walk. That count is the count of the current form, and the first
inversion is at the same value. Thus the rounding of the sum `2 + x` makes all of the
inversions, and the rounding of the quotient makes none. Correction B corrects the
rounding of the sum.

## How other libraries do it

Three libraries that give `log1p` have no inversion over 260000 adjacent pairs of the
`double` type, at the same 13 starting points:

| library | inversions |
| --- | --- |
| glibc | 0 |
| Boost.Math | 0 |
| MPFR at 53 bits | 0 |

MPFR calculates with a working precision which is larger than the result. Its Ziv loop
increases that precision until the result is correct. glibc and Boost.Math use a
polynomial in `x`, and they do not divide by `2 + x`. Thus none of the three libraries
has the reduction which makes this defect.

## The corrections

Correction A changes the form of the reduction. Correction B and correction C keep the
form and correct the rounding. Correction B corrects the reduction, and correction C
corrects the sum of the result. This report writes that pair as B and C.

Correction B and correction C are not exchangeable. Correction B alone gives the correct
order. Correction C alone does not. Correction C needs a reduction which keeps the order
of `x`, and correction A and correction B each give one.

A step of Newton at the result cannot work. Its residual holds about one digit only. The
step which works is at the reduction, where the residual is exact.

### Correction A: the reciprocal form

The identity `x / (2 + x) == 1 / (1 + 2/x)` puts `x` in one place only:

```cpp
// log1p(x) = 2 * atanh(w), with w = x / (2 + x) == 1 / (1 + 2/x).
const T w { one / (one + T { 2, 0 } / x) };

result = w * detail::log1p_series_expansion(w * w);
```

The calculation is a chain of three steps:
- The value of `2/x` decreases when `x` increases.
- The value of `1 + 2/x` then decreases.
- The value of `1 / (1 + 2/x)` then increases.

Each step moves in one direction only, and the rounding of a step keeps that direction.
Thus `w` cannot decrease when `x` increases, at every precision.

The value of `2/x` becomes larger when `|x|` becomes smaller, thus the form needs a
check of the small end. Two guards keep the reduction away from a small value of `x`:
- The first guard gives `x` for an argument which is not normal.
- The second guard gives `x` for `|x| < 1e-digits10`.

Thus the smallest argument that gets to the reduction is `1e-digits10`, and the largest
value of `2/x` is `2 * 10^digits10`. For the `decimal128_t` type that value is 2e34, and
the largest finite value of the type is about 1e6145.

A subnormal argument makes `2/x` infinite, and the guards keep it away from the
reduction. This is true also with `BOOST_DECIMAL_FAST_MATH`, which removes the first
guard. The second guard is enough:

```
decimal32_t  (small_argument = 1e-7)
  smallest subnormal     x = 1.000000e-101   2/x = inf            reduction no   log1p = 1.000000e-101
  smallest normal        x = 1.000000e-95    2/x = 2.000000e+95   reduction no   log1p = 1.000000e-95
  small_argument / 10    x = 1.000000e-08    2/x = 2.000000e+08   reduction no   log1p = 1.000000e-08
  small_argument         x = 1.000000e-07    2/x = 2.000000e+07   reduction yes  log1p = 1.000000e-07
  -smallest subnormal    x = -1.000000e-101  2/x = inf            reduction no   log1p = -1.000000e-101
```

At the smallest argument that gets to the reduction, the value of `2/x` is large. The
addition of 1 then has no effect on it. This makes no error. The term which the 1 gives
is less than one tenth of a unit of the last place. Over 200000 consecutive values from
1e-7 the largest error of the `decimal32_t` type is 0.000 eps for the current form and
for the reciprocal form.

### Correction B: the residual of the division

A step of Newton at the reduction works, because the sum and the quotient each have an
error which is available exactly:

- The value `d - 2` is exact, because `d` is between 1 and 4. Thus `e = x - (d - 2)` is
  the exact error of the sum, and `2 + x == d + e`.
- The residual of a division which rounds to the nearest value is representable in the
  same type. One `unchecked_fma` gives it.

```cpp
// log1p(x) = 2 * atanh(w), with w = x / (2 + x).
// The sum 2 + x is not exact. The value e holds the error of that sum, and
// the value r holds the residual of the quotient. The two together give w
// with the correct rounding, thus w keeps the order of x.
constexpr T two { 2, 0 };

const T d  { two + x };
const T e  { x - (d - two) };  // 2 + x == d + e, and this is exact
const T w0 { x / d };
const T r  { unchecked_fma(-w0, d, x) - w0 * e };
const T w  { w0 + r / d };

result = w * detail::log1p_series_expansion(w * w);
```

A test of 20000 values gives the count of the values for which `w` has the correct
rounding of the exact value of `x / (2 + x)`. The comparison of the corrections holds
that table. Correction B gives 100.00 percent for `decimal32_t` and for `decimal64_t`,
against 59.60 percent and 59.03 percent for the current form.

The correct rounding of a function which increases gives a value which does not
decrease. Thus the order of correction B is a property of its form. It does not depend
on the count of the values of a test. Correction A gives the correct order for a
different reason, and its value of `w` is wrong by as much as 5 ulp.

The term `w0 * e` is necessary. Without that term the code refines the quotient `x / d`,
and `d` is the rounded sum. The full walk then gives 125402 inversions, which is the
count of the current form, and the first inversion is at `x = 5.000001e-07`. Thus the
rounding of the sum `2 + x` is the whole cause of this defect, and the rounding of the
division has no part in it.

The `unchecked_fma` of `decimal32_t` and of `decimal64_t` rounds one time. The
`d128_fma_impl` function calculates `x * y + z`, and it rounds two times. Thus
correction B gives the correct rounding for `decimal128_t` at 86.38 percent of the
values only, with an error of not more than one ulp. Two ways give a better product for
that type:

- Make `d128_fma_impl` fuse the two operations. One code then applies to all three
  types. That change is not part of this report.
- Get the product with the split of Veltkamp, and add the parts. The comparison gives
  100.00 percent for all three types, thus this is enough for the reduction. It is not a
  fused operation, because the last sum of the parts rounds again.

### Correction C: the sum of the result

Correction B gives a value of `w` which is correct at every digit. The result is still
wrong by as much as 4 units of the last place. Thus the rest of the error is in the
series and in the last multiplication.

The first coefficient of the series is exactly 2 for all six types. Thus the result is
the sum of a large term and a small term:

```
2*atanh(w) = 2*w + w^3 * P(w*w)
```

The current code calculates `w * S(w*w)`. The series rounds at the size of the result,
and the product rounds again. Correction C keeps the reduction as the pair `wh + wl`,
and it adds the first term separately:

```cpp
constexpr T two { 2, 0 };

const T d  { two + x };
const T e  { x - (d - two) };  // 2 + x == d + e, and this is exact
const T wh { x / d };
const T r  { unchecked_fma(-wh, d, x) - wh * e };
const T wl { r / d };          // wh + wl holds about twice the count of digits

const T w    { wh + wl };
const T y    { w * w };
const T tail { w * y * detail::log1p_series_tail(y) };

result = unchecked_fma(two, wh, unchecked_fma(two, wl, tail));
```

The value `tail` is not more than about one part in 20 of the result, thus its rounding
is small against the result. Only the last `unchecked_fma` rounds at the size of the
result. The `log1p_series_tail` function is the same Horner loop as
`log1p_series_expansion`, over the coefficients after the first.

The comparison of the corrections gives the numbers of the pair B and C. For the 32 and
64 bit types the mean error goes to 0.000 eps, against 0.122 and 0.157 eps for
correction B alone. The largest error of the sample goes to one unit of the last place.
For `decimal128_t` the pair needs a product from a split to give those numbers, and it
then gives no error at the sample.

Correction C alone does not correct the order. With the current reduction it gives
125402 inversions of the full walk, which is the count of the current code. The section
on the full walk holds that row. Correction C makes the error smaller, and a smaller
error does not give a better order. Correction C needs a reduction which keeps the order
of `x`.

#### Correction C with correction A

Correction A also gives a reduction which keeps the order of `x`. Its chain gives one
value of `w` only, thus one residual of the last division makes the pair `wh + wl`:

```cpp
constexpr T one { 1, 0 };
constexpr T two { 2, 0 };

const T s  { one + two / x };
const T wh { one / s };
const T r  { unchecked_fma(-wh, s, one) };
const T wl { r / s };

const T w    { wh + wl };
const T y    { w * w };
const T tail { w * y * detail::log1p_series_tail(y) };

result = unchecked_fma(two, wh, unchecked_fma(two, wl, tail));
```

The pair A and C makes the mean error of correction A smaller by one half. For
`decimal32_t` it raises the count of the correct roundings from 39.6 percent to 59.2
percent. It does not give the accuracy of the pair B and C, which is 99.9 percent. The
value of `w` is still wrong by as much as 5 units of the last place. The residual of the
last division does not correct the rounding of `2/x` or of `1 + 2/x`. The table of the
reduction gives the same 68.41 percent for correction A with that residual and without
it.

The pair A and C gives 0 inversions at the full walk of both sides of the `decimal32_t`
branch. It gives 0 at the 240000 pairs of each of the other five types.

Both pairs need a multiplication and an addition with one rounding. The 128 bit types do
not have that. With the library `fma` the pair A and C gives 48 inversions of the
comparison and the pair B and C gives 803. With a product from a split each of them
gives 0. Correct that defect of `d128_fma_impl` first, or hold correction C for the 32
and 64 bit types only.

## Comparison of the corrections

One program gives every table of this section. It uses the 18 starting points and the
2000 values of the error. The count of the correct rounding of `w` comes from 20000
values. A second run of the program gives the same value at
every other column. Its times are different by not more than 4 points of percent.

A count of the inversions from 18 starting points gives a lower limit only. The section
on the full walk counts every pair of the branch.

| type | correction | inversions | max ulp | max eps | mean eps | correctly rounded | ns |
| --- | --- | --- | --- | --- | --- | --- | --- |
| `decimal32_t` | current | 1207 | +5 | 1.00 | 0.176 | 37.0% | 393 |
| `decimal32_t` | **A** | 0 | -5 | 0.99 | 0.159 | 39.6% | 403 |
| `decimal32_t` | B | 0 | +3 | 1.00 | 0.122 | 48.2% | 554 |
| `decimal32_t` | A and C | 0 | -3 | 0.99 | 0.088 | 59.2% | 552 |
| `decimal32_t` | **B and C** | 0 | -1 | 0.23 | 0.000 | 99.9% | 653 |
| `decimal32_t` | Remez | 1207 | +5 | 1.00 | 0.176 | 37.0% | 349 |
| `decimal32_t` | A, Remez | 0 | -5 | 0.99 | 0.159 | 39.6% | 351 |
| `decimal32_t` | B and C, Remez | 0 | -1 | 0.23 | 0.000 | 99.9% | 592 |
| `decimal64_t` | current | 1196 | -5 | 0.97 | 0.184 | 37.1% | 2380 |
| `decimal64_t` | **A** | 0 | -6 | 0.96 | 0.181 | 37.0% | 2418 |
| `decimal64_t` | B | 0 | -4 | 1.00 | 0.157 | 40.0% | 2624 |
| `decimal64_t` | A and C | 0 | -3 | 0.99 | 0.089 | 58.7% | 3124 |
| `decimal64_t` | **B and C** | 0 | -1 | 0.18 | 0.000 | 99.8% | 2812 |
| `decimal64_t` | Remez | 1196 | -5 | 0.97 | 0.184 | 37.1% | 1818 |
| `decimal64_t` | A, Remez | 0 | -6 | 0.96 | 0.181 | 37.0% | 1849 |
| `decimal64_t` | B and C, Remez | 0 | -1 | 0.18 | 0.000 | 99.8% | 2250 |
| `decimal128_t` | current | 1211 | +5 | 1.00 | 0.195 | 34.5% | 8311 |
| `decimal128_t` | **A** | 0 | +6 | 1.00 | 0.185 | 35.7% | 8340 |
| `decimal128_t` | B | 0 | -3 | 1.00 | 0.162 | 38.7% | 8897 |
| `decimal128_t` | B, split | 0 | -3 | 1.00 | 0.156 | 39.5% | 10055 |
| `decimal128_t` | A and C | 48 | +4 | 1.06 | 0.126 | 50.0% | 8868 |
| `decimal128_t` | A and C, split | 0 | -3 | 1.00 | 0.089 | 59.4% | 13196 |
| `decimal128_t` | B and C | 803 | -1 | 1.00 | 0.072 | 73.9% | 9383 |
| `decimal128_t` | **B and C, split** | 0 | +0 | 0.00 | 0.000 | 100.0% | 13329 |
| `decimal128_t` | Remez | 1211 | +5 | 1.00 | 0.195 | 34.5% | 5855 |
| `decimal128_t` | A, Remez | 0 | +6 | 1.00 | 0.185 | 35.7% | 5888 |
| `decimal128_t` | B and C, Remez | 803 | -1 | 1.00 | 0.073 | 73.8% | 6884 |
| `decimal128_t` | B and C, Remez, split | 0 | -1 | 0.22 | 0.000 | 99.9% | 10818 |

A row marked `split` gets the product from the split of Veltkamp. The other rows of
`decimal128_t` use the library `fma`.

A row marked `Remez` changes the polynomial of the series. The current polynomial of
Taylor has 7, 16 and 34 terms. A polynomial of Remez gives the same truncation error
with 6, 12 and 23 terms. Its first coefficient stays exactly 2, thus correction C keeps
its exact leading term. The section on the polynomial of the series gives these counts.

The two units give a different order, and the largest value of each is at a different
value of `x`. Each row with 1.00 eps is an error of one unit of the last place at a
result which starts with 1. The largest count of units of the last place is at a result
which starts with 8 or 9.

| type | correction | largest eps at x = | largest ulp at x = |
| --- | --- | --- | --- |
| `decimal32_t` | current | -9.516796e-02 | -8.234043e-02 |
| `decimal32_t` | **A** | -1.008950e-06 | 9.024446e-03 |
| `decimal32_t` | B | -9.516796e-02 | 9.738000e-03 |
| `decimal32_t` | A and C | -1.008950e-06 | 9.526579e-06 |
| `decimal32_t` | **B and C** | -3.512694e-01 | -3.512694e-01 |
| `decimal64_t` | current | -9.837186335184886e-02 | 9.633033748239520e-03 |
| `decimal64_t` | **A** | 2.089447260873834e-03 | -9.372017831754572e-02 |
| `decimal64_t` | B | 2.004749578739530e-05 | -9.075149056593017e-06 |
| `decimal64_t` | A and C | 1.008645510949761e-06 | 9.590544781648461e-05 |
| `decimal64_t` | **B and C** | -4.212600671232678e-01 | -4.719881294565426e-01 |
| `decimal128_t` | current | -9.963355914907778443442704641678252e-03 | -7.496537910527110973452423405000450e-02 |
| `decimal128_t` | **A** | -1.001713445515752603756098298370731e-05 | 9.874396988960581436570412080666685e-06 |
| `decimal128_t` | B | -9.963355914907778443442704641678252e-03 | 9.209527179976226994427331495257727e-02 |
| `decimal128_t` | A and C | 2.076848157948359307075534930116413e-01 | 9.874396988960581436570412080666685e-06 |
| `decimal128_t` | A and C, split | -1.001713445515752603756098298370731e-05 | -9.408736888556632890770351225684401e-02 |
| `decimal128_t` | **B and C, split** | - | - |

The last row has no value of `x`, because that form gives the correctly rounded result
at every value of the sample.

The reduction has its own measurement. This table covers the forms which the polynomial
of Taylor gives. It gives the count of the values for which `w` has the correct rounding
of the exact value of `x / (2 + x)`:

| type | form of w | w is correctly rounded | max ulp of w |
| --- | --- | --- | --- |
| `decimal32_t` | current | 59.60% | +3 |
| `decimal32_t` | A | 68.41% | -5 |
| `decimal32_t` | A, with a residual | 68.41% | -5 |
| `decimal32_t` | **B** | 100.00% | +0 |
| `decimal32_t` | B, split | 100.00% | +0 |
| `decimal64_t` | current | 59.03% | -3 |
| `decimal64_t` | A | 68.07% | -5 |
| `decimal64_t` | A, with a residual | 68.07% | -5 |
| `decimal64_t` | **B** | 100.00% | +0 |
| `decimal64_t` | B, split | 100.00% | +0 |
| `decimal128_t` | current | 58.84% | -3 |
| `decimal128_t` | A | 68.42% | -5 |
| `decimal128_t` | A, with a residual | 68.38% | -5 |
| `decimal128_t` | B | 86.38% | -1 |
| `decimal128_t` | **B, split** | 100.00% | +0 |

Correction B gives the correct rounding of `w` at every value of the test for the 32 and
64 bit types. The third row of each type adds the residual of the last division to the
chain of correction A. For the 32 and 64 bit types that step gives the same count as
correction A itself. For `decimal128_t` it gives a smaller count.

The error of that `w` is in the rounding of `2/x` and of `1 + 2/x`. The residual of the
last division corrects neither of them. Only correction B makes the reduction correct.

### What each correction gives

Correction A removes every inversion of all three types. It costs 3 percent for
`decimal32_t`, 2 percent for `decimal64_t` and less than 1 percent for `decimal128_t`.
It needs one line, and it needs no new type and no new table.

Correction B removes every inversion of all three types, and it also makes the error
smaller. The appendix calls a reduction in the next larger type correction D. For
`decimal32_t` and for `decimal64_t` the numbers of correction B are equal at every
column to the numbers of correction D. Thus correction B gives the accuracy of a
calculation in a larger type, with no larger type. This is of most value for
`decimal128_t`, which has no larger decimal type.

Correction C makes the error smaller again, and it needs a reduction which keeps the
order of `x`. With correction B it gives the correct rounding at 99.9 percent of the
sample for `decimal32_t`, and 99.8 percent for `decimal64_t`. The mean error goes to
0.000 eps. For `decimal128_t` it gives 73.9 percent with the library `fma`, and 100.0
percent with a product from a split. With correction A it gives 59.2 percent, 58.7
percent and 59.4 percent.

### The cost

This table gives the time of each form in `ns`, and the change against the current code:

| form | `decimal32_t` | `decimal64_t` | `decimal128_t` |
| --- | --- | --- | --- |
| current | 393 | 2380 | 8311 |
| A | 403 (+3%) | 2418 (+2%) | 8340 (+0%) |
| B | 554 (+41%) | 2624 (+10%) | 8897 (+7%) |
| A and C | 552 (+40%) | 3124 (+31%) | 8868 (+7%) |
| B and C | 653 (+66%) | 2812 (+18%) | 9383 (+13%) |
| B and C, split | - | - | 13329 (+60%) |
| Remez | 349 (-11%) | 1818 (-24%) | 5855 (-30%) |
| A, Remez | 351 (-11%) | 1849 (-22%) | 5888 (-29%) |
| B and C, Remez | 592 (+51%) | 2250 (-5%) | 6884 (-17%) |
| B and C, Remez, split | - | - | 10818 (+30%) |

The polynomial of Remez saves 1 step of Horner for `decimal32_t`, 4 steps for
`decimal64_t` and 11 steps for `decimal128_t`. It makes the current code faster by 11
percent, 24 percent and 30 percent. It gives the same value at every column of accuracy
for `decimal32_t` and for `decimal64_t`. For `decimal128_t` the rows of the pair B and C
are not equal. With a product from a split that pair gives 99.9 percent in place of
100.0 percent.

The polynomial pays for the correction. With it, correction A is faster than the current
code by 11 percent, 22 percent and 29 percent. The pair B and C then costs 51 percent
for `decimal32_t`, and it is faster than the current code by 5 percent for
`decimal64_t`. For `decimal128_t` it is faster by 17 percent with the library `fma`, and
it costs 30 percent with a product from a split.

### The choice

Use the pair B and C for the accuracy, or correction A for the smallest change and the
smallest cost. Correction A gives 0 inversions at every one of the 116000000 pairs of
the full walk. The pair A and C and the pair B and C each give 0 there. A change of the
polynomial is separate from the correction of the order, and each one is of use without
the other.

Correction A does not make the error smaller. Its largest error is 5 to 6 units of the
last place, and the largest error of the current form is 5. Correction A gives the
correct order for a different reason. The value of `w` keeps the order of `x` at every
step, thus the result keeps the order of `x`. The error can be large, and it changes in
one direction only. A correct order does not need a small error.

## The full walk

A walk of every representable value of the branch of the `decimal32_t` type gives the
count of the inversions.

| form | positive side | negative side |
| --- | --- | --- |
| current, `x / (2 + x)` | 125402 | 0 |
| **A, `1 / (1 + 2/x)`** | **0** | **0** |
| **B, the residual of the division** | **0** | **0** |
| B, without the term `w0 * e` | 125402 | 0 |
| C alone, with the current reduction | 125402 | 0 |
| **A and C** | **0** | **0** |
| **B and C** | **0** | **0** |

The first inversion of the current form is at `x = 5.000001e-07`.

Correction B without the term `w0 * e` gives the count of the current form, and the same
first inversion. That form corrects the rounding of the division only. Thus the rounding
of the sum `2 + x` is the whole cause, and the rounding of the division has no part.

Correction C alone gives the same count and the same first inversion. That correction
makes the error smaller, and it does not correct the order. Correction B and correction
C are not exchangeable. Correction B is the one which corrects the order, and correction
C needs it.

The full walk is necessary. Correction H of the appendix gives 0 inversions at the 18
starting points of the comparison, and 524 at the full walk. Those 18 points do not
touch the values at which correction H inverts. A count from a set of starting points
gives a lower limit only. Use the full walk to accept a correction.

<details>
<summary>Show the program of the full walk</summary>

```cpp
// Exhaustive decimal32_t walk of every form of this report.
#include <boost/decimal.hpp>
#include <boost/decimal/detail/cmath/impl/log1p_impl.hpp>
#include <boost/decimal/detail/cmath/fma.hpp>
#include <mpfr.h>
#include <cstdio>
#include <cstring>
#include <limits>
#include <map>
#include <string>
using D = decimal32_t;

static std::string to_scientific_string(D v)
{
    char b[200];
    auto r = to_chars(b, b + 198, v, chars_format::scientific, std::numeric_limits<D>::digits10 - 1);
    *r.ptr = 0;
    return std::string(b);
}

static D from_scientific_string(const char* p)
{
    D o {};
    from_chars(p, p + std::strlen(p), o);
    return o;
}

static D next_up(D x)
{
    return nextafter(x, std::numeric_limits<D>::infinity());
}

static D log1p_series_tail(D y)
{
    auto& c = detail::log1p_table::d32_coeffs;
    const std::size_t N = c.size();
    D r = c[N - 1];
    for (std::size_t i = N - 1; i-- > 1;)
    {
        r = detail::unchecked_fma(r, y, c[i]);
    }
    return r;
}

static D log1p_series(D w)
{
    return w * detail::log1p_series_expansion<D>(w * w);
}

static D log1p_current(D x)
{
    return log1p_series(x / (D { 2, 0 } + x));
}

static D log1p_correction_b(D x)
{
    const D tw { 2, 0 };
    const D d = tw + x, e = x - (d - tw), w = x / d;
    const D r = detail::unchecked_fma(-w, d, x) - w * e;
    return log1p_series(w + r / d);
}

static D log1p_correction_a(D x)
{
    const D o { 1, 0 };
    return log1p_series(o / (o + D { 2, 0 } / x));
}

static D log1p_correction_b_without_term(D x) // correction B without the term w0*e
{
    const D tw { 2, 0 };
    const D d = tw + x, w = x / d;
    const D r = detail::unchecked_fma(-w, d, x);
    return log1p_series(w + r / d);
}

static D log1p_correction_c(D x)
{
    const D tw { 2, 0 };
    const D w = x / (tw + x), y = w * w;
    return detail::unchecked_fma(tw, w, w * y * log1p_series_tail(y));
}

static D log1p_correction_bc(D x)
{
    const D tw { 2, 0 };
    const D d = tw + x, e = x - (d - tw), wh = x / d;
    const D r = detail::unchecked_fma(-wh, d, x) - wh * e;
    const D wl = r / d;
    const D w = wh + wl, y = w * w;
    const D tail = w * y * log1p_series_tail(y);
    return detail::unchecked_fma(tw, wh, detail::unchecked_fma(tw, wl, tail));
}

static D log1p_correction_ac(D x)
{
    const D o { 1, 0 }, tw { 2, 0 };
    const D s = o + tw / x, wh = o / s;
    const D r = detail::unchecked_fma(-wh, s, o);
    const D wl = r / s;
    const D w = wh + wl, y = w * w;
    const D tail = w * y * log1p_series_tail(y);
    return detail::unchecked_fma(tw, wh, detail::unchecked_fma(tw, wl, tail));
}

static std::map<int, std::pair<D, D>> log_of_f_table;

static void build_log_table()
{
    mpfr_t f, l, h, r;
    mpfr_inits2(600, f, l, h, r, (mpfr_ptr)0);
    char b[300];
    for (int j = -50; j <= 50; ++j)
    {
        if (j == 0)
        {
            log_of_f_table[0] = { D { 0, 0 }, D { 0, 0 } };
            continue;
        }
        mpfr_set_si(f, 100 + j, MPFR_RNDN);
        mpfr_div_si(f, f, 100, MPFR_RNDN);
        mpfr_log(l, f, MPFR_RNDN);
        mpfr_sprintf(b, "%.6Re", l);
        const D hi = from_scientific_string(b);
        mpfr_set_str(h, to_scientific_string(hi).c_str(), 10, MPFR_RNDN);
        mpfr_sub(r, l, h, MPFR_RNDN);
        mpfr_sprintf(b, "%.6Re", r);
        log_of_f_table[j] = { hi, from_scientific_string(b) };
    }
    mpfr_clears(f, l, h, r, (mpfr_ptr)0);
}

static D log1p_table_reduction(D x)
{
    const D hund { 100, 0 };
    long j = static_cast<long>(static_cast<double>(x * hund) + (x < D { 0, 0 } ? -0.5 : 0.5));
    if (j < -50)
    {
        j = -50;
    }
    if (j > 50)
    {
        j = 50;
    }
    if (j == 0)
    {
        return log1p_correction_bc(x);
    }
    const D F = D { 1, 0 } + D { static_cast<int>(j), 0 } / hund;
    const D f = x - (D { static_cast<int>(j), 0 } / hund);
    const auto& e = log_of_f_table[static_cast<int>(j)];
    return e.first + (e.second + log1p_correction_bc(f / F));
}

static void sweep(const char* side, D lo, D hi, D (*f)(D), const char* fn)
{
    long long n = 0, inv = 0, ties = 0;
    std::string first;
    D x = lo, p = f(x);
    while (x < hi)
    {
        x = next_up(x);
        ++n;
        const D v = f(x);
        if (v < p)
        {
            if (!inv)
            {
                first = to_scientific_string(x);
            }
            ++inv;
        }
        else if (v == p)
        {
            ++ties;
        }
        p = v;
    }
    std::printf("  %-9s %-24s %12lld values  inversions %-9lld ties %-11lld %s\n", side, fn, n, inv, ties,
                inv ? ("first at " + first).c_str() : "");
    std::fflush(stdout);
}

int main()
{
    build_log_table();
    std::printf("decimal32_t, every representable value of the branch\n");

    struct
    {
        const char* n;
        D (*f)(D);
    } fs[] = { { "current", log1p_current },
               { "A, reciprocal", log1p_correction_a },
               { "B, residual", log1p_correction_b },
               { "B, without w0*e", log1p_correction_b_without_term },
               { "C, leading split", log1p_correction_c },
               { "A and C", log1p_correction_ac },
               { "B and C", log1p_correction_bc },
               { "table of log(F), B, C", log1p_table_reduction } };

    std::printf("  positive side, 1e-7 .. 0.5\n");
    for (auto& e : fs)
    {
        sweep("positive", from_scientific_string("1e-7"), from_scientific_string("0.5"), e.f, e.n);
    }
    std::printf("  negative side, -0.5 .. -1e-7\n");
    for (auto& e : fs)
    {
        sweep("negative", from_scientific_string("-0.5"), from_scientific_string("-1e-7"), e.f, e.n);
    }
    std::printf("DONE\n");
    return 0;
}
```

</details>

### Output of the full walk

```
decimal32_t, every representable value of the branch
  positive side, 1e-7 .. 0.5
  positive  current                      58000000 values  inversions 125402    ties 25938889    first at 5.000001e-07
  positive  A, reciprocal                58000000 values  inversions 0         ties 31583474    
  positive  B, residual                  58000000 values  inversions 0         ties 26205491    
  positive  B, without w0*e              58000000 values  inversions 125402    ties 25938889    first at 5.000001e-07
  positive  C, leading split             58000000 values  inversions 125402    ties 25938032    first at 5.000001e-07
  positive  A and C                      58000000 values  inversions 0         ties 22149589    
  positive  B and C                      58000000 values  inversions 0         ties 1412270     
  positive  table of log(F), B, C        58000000 values  inversions 0         ties 1639604     
  negative side, -0.5 .. -1e-7
  negative  current                      58000000 values  inversions 0         ties 25187536    
  negative  A, reciprocal                58000000 values  inversions 0         ties 30305665    
  negative  B, residual                  58000000 values  inversions 0         ties 24815515    
  negative  B, without w0*e              58000000 values  inversions 0         ties 25187536    
  negative  C, leading split             58000000 values  inversions 0         ties 25187533    
  negative  A and C                      58000000 values  inversions 0         ties 21246887    
  negative  B and C                      58000000 values  inversions 0         ties 479935      
  negative  table of log(F), B, C        58000000 values  inversions 0         ties 685957      
```

The column `ties` counts the pairs which give the same result. A correction which makes
the error smaller gives fewer ties.

The last row of the output adds the table of `log(F)` of the paper of 1990, which the
section on published work names. It gives 0 inversions, and it needs a table of 101
pairs. The pair B and C gives the same order with no table.

Corrections A, B and C also remove every inversion of the other five types, over 240000
adjacent pairs at six starting points:

| type | current | A | B | A and C | B and C | B and C, with an exact product |
| --- | --- | --- | --- | --- | --- | --- |
| `decimal32_t` | 1242 | **0** | **0** | **0** | **0** | **0** |
| `decimal_fast32_t` | 1242 | **0** | **0** | **0** | **0** | **0** |
| `decimal64_t` | 1215 | **0** | **0** | **0** | **0** | **0** |
| `decimal_fast64_t` | 1215 | **0** | **0** | **0** | **0** | **0** |
| `decimal128_t` | 1250 | **0** | **0** | **0** | **0** | **0** |
| `decimal_fast128_t` | 1250 | **0** | **0** | **0** | **0** | **0** |

These six starting points do not show the limit of correction C on the 128 bit types.
The section on the effect on `asinh` shows it.

## The polynomial of the series

The series is a Taylor polynomial of `2*atanh(w)`. Two other families of polynomials
give an approximation of the same function, and a quotient of two polynomials of Pade is
a fourth choice. A better polynomial can make the error smaller, or it can make the
order correct, or it can do neither.

The value of `|x|` is not more than 1/2, thus the range of `w` is `[-1/3, 1/5]`. The
polynomial is a function of `t`, with `t = w*w`, thus the range of `t` is `[0, 1/9]`.
This table gives the smallest count of terms for an error of less than one tenth of a
unit of the last place:

| type | family | terms | largest relative error |
| --- | --- | --- | --- |
| `decimal32_t` | Taylor, as the code has it | 7 | 1.49e-08 |
| `decimal32_t` | Chebyshev | 6 | 9.92e-11 |
| `decimal32_t` | Remez | 6 | 9.86e-11 |
| `decimal32_t` | Remez, leading term held | 6 | 1.09e-10 |
| `decimal64_t` | Taylor, as the code has it | 16 | 1.76e-17 |
| `decimal64_t` | Chebyshev | 12 | 3.36e-20 |
| `decimal64_t` | Remez | 12 | 3.33e-20 |
| `decimal64_t` | Remez, leading term held | 12 | 3.50e-20 |
| `decimal128_t` | Taylor, as the code has it | 34 | 5.62e-35 |
| `decimal128_t` | Chebyshev | 23 | 2.57e-37 |
| `decimal128_t` | Remez | 23 | 2.55e-37 |
| `decimal128_t` | Remez, leading term held | 23 | 2.61e-37 |

The polynomial of Remez is the best polynomial of its degree. It gives the same count of
terms as the economization of Chebyshev, and its error is smaller by less than 1
percent. Thus 6, 12 and 23 terms is the smallest count for these three types. No other
polynomial of that degree is better.

Correction C needs the first coefficient of the series to be exactly 2. The fourth row
of each type holds that coefficient and fits the other terms. Its error is larger by 11
percent for `decimal32_t`, by 5 percent for `decimal64_t` and by 2 percent for
`decimal128_t`. The constraint costs no term. The comparison of the corrections uses
this fit for each row of Remez.
This table measures the three polynomials with the current reduction. The set of values
and the method are the same as the comparison of the corrections. The times of this
table are from one run of one program:

| type | family | inversions | max ulp | mean eps | ns |
| --- | --- | --- | --- | --- | --- |
| `decimal32_t` | Taylor | 1207 | +5 | 0.176 | 646 |
| `decimal32_t` | Chebyshev | 1207 | +5 | 0.176 | 568 |
| `decimal32_t` | Remez | 1207 | +5 | 0.176 | 565 |
| `decimal64_t` | Taylor | 1196 | -5 | 0.184 | 2632 |
| `decimal64_t` | Chebyshev | 1196 | -5 | 0.183 | 1997 |
| `decimal64_t` | Remez | 1196 | -5 | 0.183 | 2061 |
| `decimal128_t` | Taylor | 1211 | +5 | 0.195 | 9744 |
| `decimal128_t` | Chebyshev | 1211 | +5 | 0.195 | 6723 |
| `decimal128_t` | Remez | 1211 | +5 | 0.195 | 6836 |

The count of the inversions, the largest error and the mean error are equal for the
three families at all three types. The truncation error of the polynomial of Remez is
smaller by a factor of 150 than the truncation error of the current polynomial. The
result of the measurement does not change. Thus the truncation of the series is not the
cause of the error of `log1p`. The error comes from the rounding of the reduction and
from the rounding of each step of the polynomial. The polynomial of Remez is the best of
its degree, thus no other coefficients make the error smaller.

A smaller count of terms makes the function faster. From the times of this table,
Chebyshev gives 12 percent for `decimal32_t`, 24 percent for `decimal64_t` and 31
percent for `decimal128_t`. Chebyshev and Remez give the same time. Use the
economization of Chebyshev, because it comes from a closed calculation and Remez needs
an iteration.

The comparison of the corrections measures the polynomial of Remez again, in the same
program as the current code. The times of that table are not the times of this table,
because the loops are not the same.

A quotient of two polynomials of Pade was also tested, and it is not a good choice. It
has `t` in the numerator and in the denominator, thus it has the opposition of this
defect one step further in. With correction A for the reduction, which gives 0
inversions, a quotient of Pade gives 1870, 3200 and 5442 inversions for the three types.
It is also the only polynomial of this report with an error of more than one eps. This
comes from the shape of a quotient, thus no degree and no coefficients correct it. Do
not use a quotient of polynomials for this function.

## The fused multiply and add

The `taylor_series_result` function calculates one `unchecked_fma` for each coefficient.
This table changes that step to a multiplication and an addition:

| type | step of the polynomial | max ulp | mean eps | ns |
| `decimal32_t` | `unchecked_fma` | +5 | 0.176 | 646 |
| `decimal32_t` | a multiplication and an addition | +5 | 0.176 | 669 |
| `decimal64_t` | `unchecked_fma` | -5 | 0.184 | 2632 |
| `decimal64_t` | a multiplication and an addition | -5 | 0.184 | 1259 |
| `decimal128_t` | `unchecked_fma` | +5 | 0.195 | 9744 |
| `decimal128_t` | a multiplication and an addition | +5 | 0.195 | 9675 |

The `decimal64_t` type gives the same error at one half of the time. For `decimal128_t`
the two rows are equal, because `d128_fma_impl` calculates `x * y + z`. That function is
the expression of the second row.

A test of 20000 random triples gives 83.86 percent of correct results for the `fma` of
`decimal128_t`. The largest error there is 118 units of the last place. The 32 and 64
bit types give 100.00 percent. Correction C needs one rounding, thus `d128_fma_impl`
needs a correction first.
Other functions use the same `taylor_series_result`. A separate examination of all of
them is necessary, and this report does not do it.

## The effect on asinh

The `asinh` function calls `log1p` for `fourth_root_epsilon <= |x| < 0.5`. Its
inversions in that range come from `log1p`. This table walks 40000 consecutive values of
`x` from each of six points:

| type | current | A | B | A and C | B and C | B and C, with an exact product |
| --- | --- | --- | --- | --- | --- | --- |
| `decimal32_t` | 1749 | **0** | **0** | **0** | **0** | **0** |
| `decimal_fast32_t` | 1749 | **0** | **0** | **0** | **0** | **0** |
| `decimal64_t` | 1746 | **0** | **0** | **0** | **0** | **0** |
| `decimal_fast64_t` | 1746 | **0** | **0** | **0** | **0** | **0** |
| `decimal128_t` | 1768 | **0** | **0** | **0** | 232 | **0** |
| `decimal_fast128_t` | 1768 | **0** | **0** | **0** | 232 | **0** |

Correction A and correction B each remove every inversion of `asinh` in this range, for
all six types. The counts of the current code are larger than the counts of `log1p`,
because `asinh` transforms `x` before the call. The two walks give `log1p` a different
set of arguments.

The pair B and C gives 232 inversions for the 128 bit types. Correction C needs one
rounding for a multiplication and an addition, and `d128_fma_impl` gives two. The last
column gets that product from the split of Veltkamp, and it gives 0 for all six types.

The pair A and C gives 0 at these points, and it does not remove that dependence. It
gives 48 inversions of `log1p` on `decimal128_t` with the library `fma`, and 0 with a
product from a split.

This is the strongest test of this report. The walk of `log1p` alone gives 0 for all six
types at these starting points, and `asinh` finds the limit. Use `asinh` to accept
correction C.

## Published work

A paper of 1990 gives the same reduction and the same two corrections, for binary
arithmetic. Its Procedure 2 uses the transformation `a = 2f/(2 + f)`, which is `2*w` of
this report. That paper gives a bound of 0.57 units of the last place, and it does not
examine the order of the results. One unit of the last place is enough to make an
inversion in this range.

- P. T. P. Tang, "Table-driven implementation of the logarithm function in IEEE
  floating-point arithmetic", ACM Transactions on Mathematical Software, volume 16,
  number 4, pages 378-400, December 1990. Procedure 2, sections 6.2, 7.2.1 and 7.2.3.
- S. Boldo and M. Daumas, "Representable correcting terms for possibly underflowing
  floating point operations", 16th IEEE Symposium on Computer Arithmetic, 2003, pages
  79-86.
- S. Boldo and J.-M. Muller, "Some functions computable with a fused-mac", 17th IEEE
  Symposium on Computer Arithmetic, 2005.
- P. H. Sterbenz, "Floating-Point Computation", Prentice-Hall, 1974.
- S. Linnainmaa, "Software for doubled-precision floating-point computations", ACM
  Transactions on Mathematical Software, volume 7, number 3, pages 272-283, 1981.

## Appendix: the other corrections which were tested

Five other corrections were tested, and none of them is a better choice:
- Correction D, the reduction in the next larger type. It gives 0 inversions at the full
  walk, and it needs a type which is larger than the result. Thus it does not apply to
  `decimal128_t`. Correction B gives the same numbers with no larger type.
- Correction E, the whole branch in the next larger type. It gives the correct rounding
  at every value of the sample, at a cost of a factor of 3.5 to 6.3. It does not apply
  to `decimal128_t`.
- Correction F, a compensation of the residual. It makes the mean error smaller, and it
  makes the count of the inversions larger by a factor of 3.7.
- Correction G, one step of the method of Newton at the result. It makes the count of
  the inversions 2.6 times the count of the current form. It also puts 2626 inversions
  into the negative side, which has 0 inversions in the current code.
- Correction H, a polynomial in `x` with no reduction. It gives 0 inversions at the 18
  starting points, and 524 at the full walk. The product `x * P(x)` has no guarantee,
  because `P` decreases when `x` increases.

ibmibmibm:
The reduction w = x / (2 + x) rounds the sum 2 + x, and the error of that sum makes the quotient fall where x rises. The series then rounds at the size of 2, which puts the full size of the result through that rounding. Together these give pairs of adjacent arguments whose results are in the opposite order.

Two corrections remove both. The value e holds the error of the sum, and wh + wl holds the reduction to about two times the digits of the type. The first coefficient of the series is exactly 2 for every type, thus the tables now leave it out and the result adds 2*wh and 2*wl with a fused operation. Only the small value tail rounds at a small size.

test_log1p.cpp gets test_log1p_monotonic, which walks the same pairs for all six types.

Fix #1442

### DIFF
--- a/include/boost/decimal/detail/cmath/impl/log1p_impl.hpp
+++ b/include/boost/decimal/detail/cmath/impl/log1p_impl.hpp
@@ -26,19 +26,18 @@ template <bool b>
 struct log1p_table_imp
 {
 private:
-    using d32_coeffs_t  = std::array<decimal32_t,   7>;
-    using d64_coeffs_t  = std::array<decimal64_t,  16>;
-    using d128_coeffs_t = std::array<decimal128_t, 34>;
+    using d32_coeffs_t  = std::array<decimal32_t,   6>;
+    using d64_coeffs_t  = std::array<decimal64_t,  15>;
+    using d128_coeffs_t = std::array<decimal128_t, 33>;
 
-    using d32_fast_coeffs_t  = std::array<decimal_fast32_t,   7>;
-    using d64_fast_coeffs_t  = std::array<decimal_fast64_t,  16>;
-    using d128_fast_coeffs_t = std::array<decimal_fast128_t, 34>;
+    using d32_fast_coeffs_t  = std::array<decimal_fast32_t,   6>;
+    using d64_fast_coeffs_t  = std::array<decimal_fast64_t,  15>;
+    using d128_fast_coeffs_t = std::array<decimal_fast128_t, 33>;
 
 public:
     static constexpr d32_coeffs_t d32_coeffs =
     {{
-         // Series[2*ArcTanh[w], {w, 0, 13}]
-         boost::decimal::decimal32_t { 2, 0 },                               // * w^1
+         // Series[2*ArcTanh[w], {w, 0, 13}], without the first term
          boost::decimal::decimal32_t { UINT64_C(6666666666666666667), -19 }, // * w^3
          boost::decimal::decimal32_t { UINT64_C(4000000000000000000), -19 }, // * w^5
          boost::decimal::decimal32_t { UINT64_C(2857142857142857143), -19 }, // * w^7
@@ -49,8 +48,7 @@ public:
 
     static constexpr d32_fast_coeffs_t d32_fast_coeffs =
     {{
-         // Series[2*ArcTanh[w], {w, 0, 13}]
-         boost::decimal::decimal_fast32_t { 2, 0 },                               // * w^1
+         // Series[2*ArcTanh[w], {w, 0, 13}], without the first term
          boost::decimal::decimal_fast32_t { UINT64_C(6666666666666666667), -19 }, // * w^3
          boost::decimal::decimal_fast32_t { UINT64_C(4000000000000000000), -19 }, // * w^5
          boost::decimal::decimal_fast32_t { UINT64_C(2857142857142857143), -19 }, // * w^7
@@ -61,8 +59,7 @@ public:
 
     static constexpr d64_coeffs_t d64_coeffs =
     {{
-         // Series[2*ArcTanh[w], {w, 0, 31}]
-         boost::decimal::decimal64_t { 2, 0 },                                   // * w^1
+         // Series[2*ArcTanh[w], {w, 0, 31}], without the first term
          boost::decimal::decimal64_t { UINT64_C(6666666666666666667), -19 },     // * w^3
          boost::decimal::decimal64_t { UINT64_C(4000000000000000000), -19 },     // * w^5
          boost::decimal::decimal64_t { UINT64_C(2857142857142857143), -19 },     // * w^7
@@ -82,8 +79,7 @@ public:
 
     static constexpr d64_fast_coeffs_t d64_fast_coeffs =
     {{
-         // Series[2*ArcTanh[w], {w, 0, 31}]
-         boost::decimal::decimal_fast64_t { 2, 0 },                                   // * w^1
+         // Series[2*ArcTanh[w], {w, 0, 31}], without the first term
          boost::decimal::decimal_fast64_t { UINT64_C(6666666666666666667), -19 },     // * w^3
          boost::decimal::decimal_fast64_t { UINT64_C(4000000000000000000), -19 },     // * w^5
          boost::decimal::decimal_fast64_t { UINT64_C(2857142857142857143), -19 },     // * w^7
@@ -103,8 +99,7 @@ public:
 
     static constexpr d128_coeffs_t d128_coeffs =
     {{
-         // Series[2*ArcTanh[w], {w, 0, 67}]
-         ::boost::decimal::decimal128_t { 2, 0 },                                                                                        // * w^1
+         // Series[2*ArcTanh[w], {w, 0, 67}], without the first term
          ::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(361400724161834), UINT64_C(14966504185106442923) }, -34 }, // * w^3
          ::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(216840434497100), UINT64_C(16358600140547686400) }, -34 }, // * w^5
          ::boost::decimal::decimal128_t { boost::int128::uint128_t { UINT64_C(154886024640786), UINT64_C(6414216079331332681) }, -34 },  // * w^7
@@ -142,8 +137,7 @@ public:
 
     static constexpr d128_fast_coeffs_t d128_fast_coeffs =
     {{
-         // Series[2*ArcTanh[w], {w, 0, 67}]
-         ::boost::decimal::decimal_fast128_t { 2, 0 },                                                                                        // * w^1
+         // Series[2*ArcTanh[w], {w, 0, 67}], without the first term
          ::boost::decimal::decimal_fast128_t { boost::int128::uint128_t { UINT64_C(361400724161834), UINT64_C(14966504185106442923) }, -34 }, // * w^3
          ::boost::decimal::decimal_fast128_t { boost::int128::uint128_t { UINT64_C(216840434497100), UINT64_C(16358600140547686400) }, -34 }, // * w^5
          ::boost::decimal::decimal_fast128_t { boost::int128::uint128_t { UINT64_C(154886024640786), UINT64_C(6414216079331332681) }, -34 },  // * w^7
@@ -206,41 +200,44 @@ constexpr typename log1p_table_imp<b>::d128_fast_coeffs_t log1p_table_imp<b>::d1
 
 using log1p_table = log1p_detail::log1p_table_imp<true>;
 
+// 2*atanh(w) = 2*w + (2/3)*w^3 + (2/5)*w^5 + ...
+// The first coefficient is exactly 2 for every type, thus the tables leave it out and
+// the caller adds that term itself. This gives the rest of the series, divided by w^3.
 template <BOOST_DECIMAL_DECIMAL_FLOATING_TYPE T>
-constexpr auto log1p_series_expansion(T z2) noexcept;
+constexpr auto log1p_series_tail(T z2) noexcept;
 
 template <>
-constexpr auto log1p_series_expansion<decimal32_t>(decimal32_t z2) noexcept
+constexpr auto log1p_series_tail<decimal32_t>(decimal32_t z2) noexcept
 {
     return taylor_series_result(z2, log1p_table::d32_coeffs);
 }
 
 template <>
-constexpr auto log1p_series_expansion<decimal_fast32_t>(decimal_fast32_t z2) noexcept
+constexpr auto log1p_series_tail<decimal_fast32_t>(decimal_fast32_t z2) noexcept
 {
     return taylor_series_result(z2, log1p_table::d32_fast_coeffs);
 }
 
 template <>
-constexpr auto log1p_series_expansion<decimal64_t>(decimal64_t z2) noexcept
+constexpr auto log1p_series_tail<decimal64_t>(decimal64_t z2) noexcept
 {
     return taylor_series_result(z2, log1p_table::d64_coeffs);
 }
 
 template <>
-constexpr auto log1p_series_expansion<decimal_fast64_t>(decimal_fast64_t z2) noexcept
+constexpr auto log1p_series_tail<decimal_fast64_t>(decimal_fast64_t z2) noexcept
 {
     return taylor_series_result(z2, log1p_table::d64_fast_coeffs);
 }
 
 template <>
-constexpr auto log1p_series_expansion<decimal128_t>(decimal128_t z2) noexcept
+constexpr auto log1p_series_tail<decimal128_t>(decimal128_t z2) noexcept
 {
     return taylor_series_result(z2, log1p_table::d128_coeffs);
 }
 
 template <>
-constexpr auto log1p_series_expansion<decimal_fast128_t>(decimal_fast128_t z2) noexcept
+constexpr auto log1p_series_tail<decimal_fast128_t>(decimal_fast128_t z2) noexcept
 {
     return taylor_series_result(z2, log1p_table::d128_fast_coeffs);
 }
--- a/include/boost/decimal/detail/cmath/log1p.hpp
+++ b/include/boost/decimal/detail/cmath/log1p.hpp
@@ -8,6 +8,7 @@
 
 #include <boost/decimal/fwd.hpp> // NOLINT(llvm-include-order)
 #include <boost/decimal/detail/cmath/abs.hpp>
+#include <boost/decimal/detail/cmath/fma.hpp>
 #include <boost/decimal/detail/cmath/impl/log1p_impl.hpp>
 #include <boost/decimal/detail/concepts.hpp>
 #include <boost/decimal/detail/config.hpp>
@@ -74,13 +75,27 @@ constexpr auto log1p_impl(const T x) noexcept
         }
         else
         {
-            // log1p(x) = 2 * atanh(w), with w = x / (2 + x).
-            // For |x| not more than 1/2, the value of |w| is not more than 1/3.
-            // The coefficient table covers this range. For all larger values
-            // of |x|, the first branch calculates log(x + 1).
-            const T w { x / (T { 2, 0 } + x) };
+            // log1p(x) = 2 * atanh(w), with w = x / (2 + x). For |x| not more than
+            // 1/2 the value of |w| is not more than 1/3, which the coefficient table
+            // covers. The first branch takes every larger |x| through log(x + 1).
+            // Two corrections make the result increase at every argument. The sum
+            // 2 + x is not exact, and its error makes the quotient fall where x rises,
+            // thus e holds that error and wh + wl holds the reduction to about two
+            // times the digits of the type. The first coefficient is exactly 2, thus
+            // 2*wh stays out of the rounding of the series and only tail rounds.
+            constexpr T two { 2, 0 };
 
-            result = w * detail::log1p_series_expansion(w * w);
+            const T d  { two + x };
+            const T e  { x - (d - two) };  // 2 + x == d + e, exact
+            const T wh { x / d };
+            const T r  { detail::unchecked_fma(-wh, d, x) - wh * e };
+            const T wl { r / d };
+
+            const T w    { wh + wl };
+            const T y    { w * w };
+            const T tail { w * y * detail::log1p_series_tail(y) };
+
+            result = detail::unchecked_fma(two, wh, detail::unchecked_fma(two, wl, tail));
         }
     }
 
--- a/test/test_log1p.cpp
+++ b/test/test_log1p.cpp
@@ -435,6 +435,57 @@ namespace local
     return result_is_ok;
   }
 
+  template<typename DecimalType>
+  auto test_log1p_monotonic(const char* type_name, const int pair_count) -> bool
+  {
+    using decimal_type = DecimalType;
+
+    // log1p increases at every argument, thus two adjacent values must not give
+    // results in the opposite order. Each walk stays below 0.5, the join of the two
+    // branches, which has a pair of its own that this test does not cover.
+
+    using start_array_type = std::array<decimal_type, 6U>;
+
+    const start_array_type start_values =
+    {{
+      decimal_type {  5, -2 }, decimal_type {  1, -1 }, decimal_type {  2, -1 },
+      decimal_type {  3, -1 }, decimal_type {  4, -1 }, decimal_type { 45, -2 },
+    }};
+
+    const decimal_type one { 1, 0 };
+
+    auto inversion_count = static_cast<std::uintmax_t>(UINT8_C(0));
+
+    for(auto i = static_cast<std::size_t>(UINT8_C(0)); i < std::tuple_size<start_array_type>::value; ++i)
+    {
+      decimal_type z    { start_values[i] };
+      decimal_type prev { log1p(z) };
+
+      for(auto n = 0; n < pair_count; ++n)
+      {
+        z = nextafter(z, one);
+
+        const decimal_type cur { log1p(z) };
+
+        if(cur < prev)
+        {
+          ++inversion_count;
+        }
+
+        prev = cur;
+      }
+    }
+
+    const bool result_is_ok { inversion_count == static_cast<std::uintmax_t>(UINT8_C(0)) };
+
+    if(!result_is_ok)
+    {
+      std::cout << "log1p is not monotonic for " << type_name << ": " << inversion_count << " inversions" << std::endl;
+    }
+
+    return result_is_ok;
+  }
+
 } // namespace local
 
 auto main() -> int
@@ -495,6 +546,27 @@ auto main() -> int
     BOOST_TEST(result_wide_is_ok);
 
     result_is_ok = (result_wide_is_ok && result_is_ok);
+  }
+
+  {
+    using namespace boost::decimal;
+
+    const auto result_mono_d32_is_ok   = local::test_log1p_monotonic<decimal32_t>      ("decimal32_t",       40000);
+    const auto result_mono_d64_is_ok   = local::test_log1p_monotonic<decimal64_t>      ("decimal64_t",       40000);
+    const auto result_mono_d128_is_ok  = local::test_log1p_monotonic<decimal128_t>     ("decimal128_t",       2000);
+    const auto result_mono_f32_is_ok   = local::test_log1p_monotonic<decimal_fast32_t> ("decimal_fast32_t",  40000);
+    const auto result_mono_f64_is_ok   = local::test_log1p_monotonic<decimal_fast64_t> ("decimal_fast64_t",  40000);
+    const auto result_mono_f128_is_ok  = local::test_log1p_monotonic<decimal_fast128_t>("decimal_fast128_t",  2000);
+
+    const auto result_monotonic_is_ok =
+    (
+         result_mono_d32_is_ok  && result_mono_d64_is_ok  && result_mono_d128_is_ok
+      && result_mono_f32_is_ok  && result_mono_f64_is_ok  && result_mono_f128_is_ok
+    );
+
+    BOOST_TEST(result_monotonic_is_ok);
+
+    result_is_ok = (result_monotonic_is_ok && result_is_ok);
   }
 
   result_is_ok = ((boost::report_errors() == 0) && result_is_ok);


mborland (approved):
This seems reasonable to me. It looked like GitHub was having some issues so more than half of the Drone run failed. I restarted that just now and will merge when that runs green.

PR #1445, opened 2026-09-04
Make the fma of the 128 bit types round one time

### LINKED ISSUE #1444: The fma of the 128 bit decimal types rounds two times
**TL;DR** — `fma` must calculate `x * y + z` with one rounding. IEEE 754-2019 gives that
definition for `fusedMultiplyAdd`. For the 128 bit types the library rounds two times,
because `d128_fma_impl` is the expression `x * y + z`.

- 16.14% of 20000 random triples are wrong, and the largest error is 118 ulp.
- If `z` is `-(x * y)`, the correct result is the rounding error of the product. All 5000 pairs of that test give 0.
- Four functions have the defect: `fma` and `unchecked_fma`, of `decimal128_t` and of `decimal_fast128_t`.
- The 32 and 64 bit types round one time. They are correct at every test point.
- The defect holds the last 281 inversions of `asinh` on the 128 bit types.

```cpp
template <bool, BOOST_DECIMAL_DECIMAL_FLOATING_TYPE T>
constexpr auto d128_fma_impl(T x, T y, T z) noexcept -> T
{
    return x * y + z;
}
```

**The correction** — `d128_fma_impl` holds the exact product of the two significands in
256 bits with `umul256`, it aligns the addend to that product, and the constructor rounds
one time. That gives 100.00% correct results on the same 20000 triples and 0 inversions
of `asinh`. One call costs 335 ns on that test, against 360 ns for the expression which
rounds two times. A shape which puts the addend far from the product costs more than the
expression. The section on the correction gives the widths which make 256 bits
sufficient, and the cost of every shape.

The parts that follow give the measurements, the exact values, and the programs.

<details>
<summary><b>The size of the error, and the worst triple</b></summary>

A test of 20000 random triples counts the results which are equal to the correctly
rounded value of MPFR:

| type | `unchecked_fma` is correct | largest error |
| --- | --- | --- |
| `decimal32_t` | 100.00% | 0 ulp |
| `decimal64_t` | 100.00% | 0 ulp |
| `decimal128_t` | 83.86% | 118 ulp |

The largest error is 118 units of the last place, and not one unit. The product and the
sum can cancel. The rounding of the product then moves into the leading digits of the
result. This is the worst of the 20000 triples:

```
x        = 4.891128916954585116359252097589272e+03
y        = -7.570510793333322422709743010426965e-02
z        = 3.707536042839128637930999007415480e+02
x*y+z    = 4.7016171001876124690127467051581179896737236445064e-01   (exact, 50 digits)
correct  = 4.701617100187612469012746705158118e-01
fma      = 4.701617100187612469012746705158000e-01   (-118 ulp)
```

</details>

<details>
<summary><b>Environment</b></summary>

- Boost.Decimal at commit `1297a5ef` of `develop`
- The correct values come from MPFR 4.2.2. The tests call `mpfr_fma` with 700 bits
- The comparison with `std::fma` uses GNU libc 2.44

</details>

<details>
<summary><b>The test program, and the accuracy and the cost of every type</b></summary>

The program makes 20000 triples of random values for each type. Each value has the full
count of digits of its type, and an exponent between `10^-3` and `10^3`. For each triple
the program calculates `unchecked_fma(x, y, z)`, and it compares that result with the
correctly rounded value from MPFR.

The program also measures a second form, which gets the product from the split of
Veltkamp. That form is not a fused operation, and the table shows its error. It is in
the program to show the size of the effect, and it is not the correction.

<details>

```cpp
// Is fma correctly rounded for each decimal type?
#include <boost/decimal.hpp>
#include <boost/decimal/detail/cmath/fma.hpp>
#include <mpfr.h>
#include <chrono>
#include <cstdio>
#include <cstring>
#include <limits>
#include <random>
#include <string>
#include <vector>

template <typename T>
static std::string to_scientific_string(T v)
{
    char b[400];
    auto r = to_chars(b, b + 398, v, chars_format::scientific, std::numeric_limits<T>::digits10 - 1);
    *r.ptr = 0;
    return std::string(b);
}

template <typename T>
static T from_scientific_string(const char* p)
{
    T o {};
    from_chars(p, p + std::strlen(p), o);
    return o;
}

template <typename T>
static T next_up(T x)
{
    return nextafter(x, std::numeric_limits<T>::infinity());
}

template <typename T>
static T next_down(T x)
{
    return nextafter(x, -std::numeric_limits<T>::infinity());
}

template <typename T>
static long ulp_distance(T g, T r)
{
    if (g == r)
    {
        return 0;
    }
    const bool h = g > r;
    T c = r;
    for (long i = 1; i <= 100000; ++i)
    {
        c = h ? next_up(c) : next_down(c);
        if (c == g)
        {
            return h ? i : -i;
        }
    }
    return h ? 100000 : -100000;
}

template <typename T>
static T veltkamp_constant();

template <>
decimal32_t veltkamp_constant<decimal32_t>()
{
    return decimal32_t { 10001, 0 };
}

template <>
decimal64_t veltkamp_constant<decimal64_t>()
{
    return decimal64_t { 100000001LL, 0 };
}

template <>
decimal128_t veltkamp_constant<decimal128_t>()
{
    decimal128_t v {};
    const char* p = "1.00000000000000001e17";
    from_chars(p, p + std::strlen(p), v);
    return v;
}

template <typename T>
static T fma_from_exact_product(T a, T b, T c)
{
    const T k = veltkamp_constant<T>();
    const T ta = k * a, ah = ta - (ta - a), al = a - ah;
    const T tb = k * b, bh = tb - (tb - b), bl = b - bh;
    const T p = a * b;
    const T err = ((ah * bh - p) + ah * bl + al * bh) + al * bl;
    const T sh = p + c, sl = (p - sh) + c;
    return sh + (sl + err);
}

template <typename T>
static std::vector<T> make_sample(size_t n, unsigned seed)
{
    constexpr int digits = std::numeric_limits<T>::digits10;
    std::mt19937_64 rng { seed };
    std::uniform_int_distribution<int> dig { 0, 9 }, lead { 1, 9 }, dec { -3, 3 }, sgn { 0, 1 };
    std::vector<T> v;
    while (v.size() < n)
    {
        std::string m;
        m += static_cast<char>('0' + lead(rng));
        for (int i = 1; i < digits; ++i)
        {
            m += static_cast<char>('0' + dig(rng));
        }
        T x = from_scientific_string<T>((m + "e" + std::to_string(dec(rng) - digits + 1)).c_str());
        if (sgn(rng))
        {
            x = -x;
        }
        v.push_back(x);
    }
    return v;
}

template <typename T>
static void check_fma(const char* type_name)
{
    constexpr int digits = std::numeric_limits<T>::digits10;
    const auto x_values = make_sample<T>(20000, 11u), y_values = make_sample<T>(20000, 22u),
               z_values = make_sample<T>(20000, 33u);
    mpfr_t a, b, c, r;
    mpfr_inits2(700, a, b, c, r, (mpfr_ptr)0);
    long library_wrong = 0, split_wrong = 0, library_max_ulp = 0, split_max_ulp = 0;
    std::string first_x, first_y, first_z, first_got, first_ref;
    for (size_t i = 0; i < x_values.size(); ++i)
    {
        const T x = x_values[i], y = y_values[i], z = z_values[i];
        mpfr_set_str(a, to_scientific_string(x).c_str(), 10, MPFR_RNDN);
        mpfr_set_str(b, to_scientific_string(y).c_str(), 10, MPFR_RNDN);
        mpfr_set_str(c, to_scientific_string(z).c_str(), 10, MPFR_RNDN);
        mpfr_fma(r, a, b, c, MPFR_RNDN);
        char buffer[420];
        mpfr_sprintf(buffer, "%.*Re", digits - 1, r);
        const T correct = from_scientific_string<T>(buffer);
        const T from_library = detail::unchecked_fma(x, y, z);
        const T from_split = fma_from_exact_product(x, y, z);
        if (from_library != correct)
        {
            ++library_wrong;
            const long u = ulp_distance(from_library, correct);
            if (std::labs(u) > std::labs(library_max_ulp))
            {
                library_max_ulp = u;
            }
            if (first_x.empty())
            {
                first_x = to_scientific_string(x);
                first_y = to_scientific_string(y);
                first_z = to_scientific_string(z);
                first_got = to_scientific_string(from_library);
                first_ref = to_scientific_string(correct);
            }
        }
        if (from_split != correct)
        {
            ++split_wrong;
            const long u = ulp_distance(from_split, correct);
            if (std::labs(u) > std::labs(split_max_ulp))
            {
                split_max_ulp = u;
            }
        }
    }
    mpfr_clears(a, b, c, r, (mpfr_ptr)0);
    std::printf("| `%s` | %.2f%% | %+ld | %.2f%% | %+ld |\n", type_name,
                100.0 * (x_values.size() - library_wrong) / x_values.size(), library_max_ulp,
                100.0 * (x_values.size() - split_wrong) / x_values.size(), split_max_ulp);
    if (!first_x.empty())
    {
        std::printf("<!-- %s first miss: x=%s y=%s z=%s got=%s correct=%s -->\n", type_name, first_x.c_str(),
                    first_y.c_str(), first_z.c_str(), first_got.c_str(), first_ref.c_str());
    }
    std::fflush(stdout);
}

template <typename T, typename F>
static double benchmark(F f, const std::vector<T>& x_values, const std::vector<T>& y_values,
                        const std::vector<T>& z_values)
{
    double best = 1e30;
    for (int rep = 0; rep < 7; ++rep)
    {
        auto t0 = std::chrono::steady_clock::now();
        T acc {};
        for (int k = 0; k < 20; ++k)
        {
            for (size_t i = 0; i < x_values.size(); ++i)
            {
                acc += f(x_values[i], y_values[i], z_values[i]);
            }
        }
        auto t1 = std::chrono::steady_clock::now();
        volatile double sink = static_cast<double>(acc);
        (void)sink;
        const double ns =
            std::chrono::duration<double, std::nano>(t1 - t0).count() / (20.0 * x_values.size());
        if (ns < best)
        {
            best = ns;
        }
    }
    return best;
}

int main()
{
    std::printf("fma against the correctly rounded value of MPFR, 20000 random triples per type\n\n");
    std::printf("| type | `unchecked_fma` is correct | max ulp | with an exact product | max ulp |\n");
    std::printf("| --- | --- | --- | --- | --- |\n");
    check_fma<decimal32_t>("decimal32_t");
    check_fma<decimal64_t>("decimal64_t");
    check_fma<decimal128_t>("decimal128_t");
    std::printf("\ncost of one call, ns\n\n");
    std::printf("| type | `unchecked_fma` | with an exact product |\n");
    std::printf("| --- | --- | --- |\n");
    {
        using T = decimal32_t;
        const auto x = make_sample<T>(2000, 11u), y = make_sample<T>(2000, 22u),
                   z = make_sample<T>(2000, 33u);
        std::printf("| `decimal32_t` | %.0f | %.0f |\n",
                    benchmark<T>([](T a, T b, T c) { return detail::unchecked_fma(a, b, c); }, x, y, z),
                    benchmark<T>([](T a, T b, T c) { return fma_from_exact_product(a, b, c); }, x, y, z));
    }
    {
        using T = decimal64_t;
        const auto x = make_sample<T>(2000, 11u), y = make_sample<T>(2000, 22u),
                   z = make_sample<T>(2000, 33u);
        std::printf("| `decimal64_t` | %.0f | %.0f |\n",
                    benchmark<T>([](T a, T b, T c) { return detail::unchecked_fma(a, b, c); }, x, y, z),
                    benchmark<T>([](T a, T b, T c) { return fma_from_exact_product(a, b, c); }, x, y, z));
    }
    {
        using T = decimal128_t;
        const auto x = make_sample<T>(2000, 11u), y = make_sample<T>(2000, 22u),
                   z = make_sample<T>(2000, 33u);
        std::printf("| `decimal128_t` | %.0f | %.0f |\n",
                    benchmark<T>([](T a, T b, T c) { return detail::unchecked_fma(a, b, c); }, x, y, z),
                    benchmark<T>([](T a, T b, T c) { return fma_from_exact_product(a, b, c); }, x, y, z));
    }
    std::printf("DONE\n");
    return 0;
}
```

</details>

**Output of the program**

```
fma against the correctly rounded value of MPFR, 20000 random triples per type

| type | `unchecked_fma` is correct | max ulp | with an exact product | max ulp |
| `decimal32_t` | 100.00% | +0 | 99.47% | -1 |
| `decimal64_t` | 100.00% | +0 | 99.47% | +1 |
| `decimal128_t` | 83.86% | -118 | 99.48% | +1 |

cost of one call, ns

| type | `unchecked_fma` | with an exact product |
| --- | --- | --- |
| `decimal32_t` | 83 | 340 |
| `decimal64_t` | 183 | 479 |
| `decimal128_t` | 360 | 1698 |
```

</details>

<details>
<summary><b>The three worst values, with the exact products</b></summary>

The program that follows prints the exact value of `x * y + z` at 50 digits, with
the correctly rounded result and the result of the library. It also does the same
comparison for glibc.

<details>
<summary>Show the program of the worst values</summary>

```cpp
// The worst cases of fma for decimal128_t, with the exact values, and a check of the
#include <algorithm>
// same operation in other places.
#include <boost/decimal.hpp>
#include <boost/decimal/detail/cmath/fma.hpp>
#include <mpfr.h>
#include <cmath>
#include <cstdio>
#include <cstring>
#include <limits>
#include <random>
#include <string>
#include <vector>
using T = decimal128_t;

static std::string to_scientific_string(T v)
{
    char b[400];
    auto r = to_chars(b, b + 398, v, chars_format::scientific, 33);
    *r.ptr = 0;
    return std::string(b);
}

static T from_scientific_string(const char* p)
{
    T o {};
    from_chars(p, p + std::strlen(p), o);
    return o;
}

static T next_up(T x)
{
    return nextafter(x, std::numeric_limits<T>::infinity());
}

static T next_down(T x)
{
    return nextafter(x, -std::numeric_limits<T>::infinity());
}

static long ulp_distance(T g, T r)
{
    if (g == r)
    {
        return 0;
    }
    const bool h = g > r;
    T c = r;
    for (long i = 1; i <= 400000; ++i)
    {
        c = h ? next_up(c) : next_down(c);
        if (c == g)
        {
            return h ? i : -i;
        }
    }
    return h ? 400000 : -400000;
}

static std::vector<T> make_sample(size_t n, unsigned seed)
{
    std::mt19937_64 rng { seed };
    std::uniform_int_distribution<int> dig { 0, 9 }, lead { 1, 9 }, dec { -3, 3 }, sgn { 0, 1 };
    std::vector<T> v;
    while (v.size() < n)
    {
        std::string m;
        m += static_cast<char>('0' + lead(rng));
        for (int i = 1; i < 34; ++i)
        {
            m += static_cast<char>('0' + dig(rng));
        }
        T x = from_scientific_string((m + "e" + std::to_string(dec(rng) - 33)).c_str());
        if (sgn(rng))
        {
            x = -x;
        }
        v.push_back(x);
    }
    return v;
}

int main()
{
    // 1. the same operation in other places
    std::mt19937_64 rng { 7u };
    std::uniform_real_distribution<double> ud { -1000.0, 1000.0 };
    mpfr_t a, b, c, r;
    mpfr_inits2(400, a, b, c, r, (mpfr_ptr)0);
    long okd = 0, okl = 0;
    const int N = 20000;
    for (int i = 0; i < N; ++i)
    {
        const double x = ud(rng), y = ud(rng), z = ud(rng);
        mpfr_set_d(a, x, MPFR_RNDN);
        mpfr_set_d(b, y, MPFR_RNDN);
        mpfr_set_d(c, z, MPFR_RNDN);
        mpfr_fma(r, a, b, c, MPFR_RNDN);
        if (mpfr_get_d(r, MPFR_RNDN) == std::fma(x, y, z))
        {
            ++okd;
        }
        mpfr_set_ld(a, (long double)x, MPFR_RNDN);
        mpfr_set_ld(b, (long double)y, MPFR_RNDN);
        mpfr_set_ld(c, (long double)z, MPFR_RNDN);
        mpfr_fma(r, a, b, c, MPFR_RNDN);
        if (mpfr_get_ld(r, MPFR_RNDN) == std::fma((long double)x, (long double)y, (long double)z))
        {
            ++okl;
        }
    }
    std::printf("| operation | correctly rounded of %d triples |\n| --- | --- |\n", N);
    std::printf("| `std::fma` of `double`, glibc | %.2f%% |\n", 100.0 * okd / N);
    std::printf("| `std::fma` of `long double`, glibc | %.2f%% |\n", 100.0 * okl / N);

    // 2. the worst cases of decimal128_t, with the exact values
    const auto xs = make_sample(20000, 11u), ys = make_sample(20000, 22u), zs = make_sample(20000, 33u);

    struct bad_case
    {
        long u;
        std::string x, y, z, got, ref, exact;
    };

    std::vector<bad_case> bad;
    for (size_t i = 0; i < xs.size(); ++i)
    {
        const T x = xs[i], y = ys[i], z = zs[i];
        mpfr_set_str(a, to_scientific_string(x).c_str(), 10, MPFR_RNDN);
        mpfr_set_str(b, to_scientific_string(y).c_str(), 10, MPFR_RNDN);
        mpfr_set_str(c, to_scientific_string(z).c_str(), 10, MPFR_RNDN);
        mpfr_fma(r, a, b, c, MPFR_RNDN);
        char q[500];
        mpfr_sprintf(q, "%.33Re", r);
        const T ref = from_scientific_string(q);
        const T got = detail::unchecked_fma(x, y, z);
        if (got != ref)
        {
            char e2[600];
            mpfr_sprintf(e2, "%.49Re", r);
            bad.push_back({ ulp_distance(got, ref), to_scientific_string(x), to_scientific_string(y),
                            to_scientific_string(z), to_scientific_string(got), to_scientific_string(ref),
                            std::string(e2) });
        }
    }
    std::sort(bad.begin(), bad.end(),
              [](const bad_case& p, const bad_case& q) { return std::labs(p.u) > std::labs(q.u); });
    std::printf("\nthe three worst of 20000 triples, `decimal128_t`\n\n");
    for (int i = 0; i < 3 && i < (int)bad.size(); ++i)
    {
        std::printf("x        = %s\ny        = %s\nz        = %s\nx*y+z    = %s   (exact, 50 "
                    "digits)\ncorrect  = %s\nfma      = %s   (%+ld ulp)\n\n",
                    bad[i].x.c_str(), bad[i].y.c_str(), bad[i].z.c_str(), bad[i].exact.c_str(),
                    bad[i].ref.c_str(), bad[i].got.c_str(), bad[i].u);
    }
    mpfr_clears(a, b, c, r, (mpfr_ptr)0);
    std::printf("DONE\n");
    return 0;
}
```

</details>

**Output of the program**

```
| operation | correctly rounded of 20000 triples |
| --- | --- |
| `std::fma` of `double`, glibc | 100.00% |
| `std::fma` of `long double`, glibc | 100.00% |

the three worst of 20000 triples, `decimal128_t`

x        = 4.891128916954585116359252097589272e+03
y        = -7.570510793333322422709743010426965e-02
z        = 3.707536042839128637930999007415480e+02
x*y+z    = 4.7016171001876124690127467051581179896737236445064e-01   (exact, 50 digits)
correct  = 4.701617100187612469012746705158118e-01
fma      = 4.701617100187612469012746705158000e-01   (-118 ulp)

x        = 9.773367626875150075486696879175602e-02
y        = -3.337907954067831540381441089089419e-01
z        = 3.235029480287355880201961692407601e-02
x*y+z    = -2.7230673690205138073617981912911544902560992129267e-04   (exact, 50 digits)
correct  = -2.723067369020513807361798191291154e-04
fma      = -2.723067369020513807361798191291200e-04   (-46 ulp)

x        = -1.635908948849659232280107483909349e+02
y        = -3.176446739966658464358952445855542e-03
z        = -5.281416298488918666313435068678353e-01
x*y+z    = -8.5038651033135564349964438059144546425870315304617e-03   (exact, 50 digits)
correct  = -8.503865103313556434996443805914455e-03
fma      = -8.503865103313556434996443805914500e-03   (-45 ulp)

```

</details>

<details>
<summary><b>The public fma of the six types, and full cancellation</b></summary>

The first program calls `detail::unchecked_fma`. This program calls the public `fma`,
on all six types. It also does the test of full cancellation: it puts `z = -(x * y)`,
thus the correct result is the rounding error of the product. It counts the pairs where
the result is 0 and the correct value is not 0.

<details>
<summary>Show the program of the public fma and the cancellation</summary>

```cpp
// The public fma of all six types, and the case of full cancellation.
#include <boost/decimal.hpp>
#include <mpfr.h>
#include <cstdio>
#include <cstring>
#include <cstdlib>
#include <limits>
#include <random>
#include <string>
#include <vector>

template <typename T>
static std::string to_scientific_string(T v)
{
    char b[400];
    auto r = to_chars(b, b + 398, v, chars_format::scientific, std::numeric_limits<T>::digits10 - 1);
    *r.ptr = 0;
    return std::string(b);
}

template <typename T>
static T from_scientific_string(const char* p)
{
    T o {};
    from_chars(p, p + std::strlen(p), o);
    return o;
}

template <typename T>
static long ulp_distance(T g, T r)
{
    if (g == r) { return 0; }
    const bool h = g > r;
    T c = r;
    for (long i = 1; i <= 100000; ++i)
    {
        c = h ? nextafter(c, std::numeric_limits<T>::infinity())
              : nextafter(c, -std::numeric_limits<T>::infinity());
        if (c == g) { return h ? i : -i; }
    }
    return h ? 100000 : -100000;
}

template <typename T>
static std::vector<T> make_sample(size_t n, unsigned seed)
{
    constexpr int digits = std::numeric_limits<T>::digits10;
    std::mt19937_64 rng { seed };
    std::uniform_int_distribution<int> dig { 0, 9 }, lead { 1, 9 }, dec { -3, 3 }, sgn { 0, 1 };
    std::vector<T> v;
    while (v.size() < n)
    {
        std::string m;
        m += static_cast<char>('0' + lead(rng));
        for (int i = 1; i < digits; ++i) { m += static_cast<char>('0' + dig(rng)); }
        T x = from_scientific_string<T>((m + "e" + std::to_string(dec(rng) - digits + 1)).c_str());
        if (sgn(rng)) { x = -x; }
        v.push_back(x);
    }
    return v;
}

// the correctly rounded x*y+z
template <typename T>
static T reference_fma(T x, T y, T z)
{
    constexpr int digits = std::numeric_limits<T>::digits10;
    mpfr_t a, b, c, r;
    mpfr_inits2(700, a, b, c, r, (mpfr_ptr)0);
    mpfr_set_str(a, to_scientific_string(x).c_str(), 10, MPFR_RNDN);
    mpfr_set_str(b, to_scientific_string(y).c_str(), 10, MPFR_RNDN);
    mpfr_set_str(c, to_scientific_string(z).c_str(), 10, MPFR_RNDN);
    mpfr_fma(r, a, b, c, MPFR_RNDN);
    char buffer[420];
    mpfr_sprintf(buffer, "%.*Re", digits - 1, r);
    mpfr_clears(a, b, c, r, (mpfr_ptr)0);
    return from_scientific_string<T>(buffer);
}

template <typename T>
static void check_public_fma(const char* type_name)
{
    const auto xs = make_sample<T>(20000, 11u), ys = make_sample<T>(20000, 22u),
               zs = make_sample<T>(20000, 33u);
    long wrong = 0, max_ulp = 0;
    for (size_t i = 0; i < xs.size(); ++i)
    {
        const T correct = reference_fma(xs[i], ys[i], zs[i]);
        const T got = fma(xs[i], ys[i], zs[i]);
        if (got != correct)
        {
            ++wrong;
            const long u = ulp_distance(got, correct);
            if (std::labs(u) > std::labs(max_ulp)) { max_ulp = u; }
        }
    }
    std::printf("| `%s` | %.2f%% | %+ld |\n", type_name, 100.0 * (xs.size() - wrong) / xs.size(), max_ulp);
    std::fflush(stdout);
}

// z = -(x*y), thus x*y+z is the rounding error of the product alone
template <typename T>
static void check_cancellation(const char* type_name)
{
    const auto xs = make_sample<T>(5000, 11u), ys = make_sample<T>(5000, 22u);
    long wrong = 0, zero_but_not = 0, tested = 0;
    std::string ex, ey, egot, eref;
    for (size_t i = 0; i < xs.size(); ++i)
    {
        const T x = xs[i], y = ys[i], z = -(x * y);
        const T correct = reference_fma(x, y, z);
        if (correct == T { 0, 0 }) { continue; }  // the product was exact
        ++tested;
        const T got = fma(x, y, z);
        if (got != correct)
        {
            ++wrong;
            if (got == T { 0, 0 }) { ++zero_but_not; }
            if (ex.empty())
            {
                ex = to_scientific_string(x);
                ey = to_scientific_string(y);
                egot = to_scientific_string(got);
                eref = to_scientific_string(correct);
            }
        }
    }
    std::printf("| `%s` | %ld | %.2f%% | %ld |\n", type_name, tested,
                100.0 * (tested - wrong) / (tested ? tested : 1), zero_but_not);
    if (!ex.empty())
    {
        std::printf("<!-- %s: x=%s y=%s z=-(x*y) got=%s correct=%s -->\n", type_name, ex.c_str(), ey.c_str(),
                    egot.c_str(), eref.c_str());
    }
    std::fflush(stdout);
}

int main()
{
    std::printf("the public fma against MPFR, 20000 random triples per type\n\n");
    std::printf("| type | `fma` is correct | max ulp |\n| --- | --- | --- |\n");
    check_public_fma<decimal32_t>("decimal32_t");
    check_public_fma<decimal_fast32_t>("decimal_fast32_t");
    check_public_fma<decimal64_t>("decimal64_t");
    check_public_fma<decimal_fast64_t>("decimal_fast64_t");
    check_public_fma<decimal128_t>("decimal128_t");
    check_public_fma<decimal_fast128_t>("decimal_fast128_t");

    std::printf("\nfull cancellation, z = -(x*y), 5000 pairs per type\n\n");
    std::printf("| type | pairs with an inexact product | `fma` is correct | result is 0 but the correct "
                "value is not |\n| --- | --- | --- | --- |\n");
    check_cancellation<decimal32_t>("decimal32_t");
    check_cancellation<decimal_fast32_t>("decimal_fast32_t");
    check_cancellation<decimal64_t>("decimal64_t");
    check_cancellation<decimal_fast64_t>("decimal_fast64_t");
    check_cancellation<decimal128_t>("decimal128_t");
    check_cancellation<decimal_fast128_t>("decimal_fast128_t");
    std::printf("DONE\n");
}
```

</details>

**Output of the program**

```
the public fma against MPFR, 20000 random triples per type

| type | `fma` is correct | max ulp |
| --- | --- | --- |
| `decimal32_t` | 100.00% | +0 |
| `decimal_fast32_t` | 100.00% | +0 |
| `decimal64_t` | 100.00% | +0 |
| `decimal_fast64_t` | 100.00% | +0 |
| `decimal128_t` | 83.86% | -118 |
| `decimal_fast128_t` | 83.86% | -118 |

full cancellation, z = -(x*y), 5000 pairs per type

| type | pairs with an inexact product | `fma` is correct | result is 0 but the correct value is not |
| --- | --- | --- | --- |
| `decimal32_t` | 5000 | 100.00% | 0 |
| `decimal_fast32_t` | 5000 | 100.00% | 0 |
| `decimal64_t` | 5000 | 100.00% | 0 |
| `decimal_fast64_t` | 5000 | 100.00% | 0 |
| `decimal128_t` | 5000 | 0.00% | 5000 |
| `decimal_fast128_t` | 5000 | 0.00% | 5000 |
```

This is the first pair of the cancellation test:

```
x       = 2.736028658208323308939783440064889e-02
y       = -8.404866370008119216028549045310085e+02
z       = -(x * y)
correct = -3.494628038600780284066993526105565e-33
fma     = -0.000000000000000000000000000000000e+00
```

</details>

<details>
<summary><b>The correction, and what it costs</b></summary>

`d128_fma_impl` holds the exact product of the two significands in 256 bits, it aligns
the addend to that product, and it rounds one time at the end. `umul256` of
`detail/u256.hpp` gives the product, and the constructor gives the rounding.

`d128_add_impl_new` cannot take the two operands as they are. Its promotion type is
already `u256`, and its `max_shift` is `digits10(u256) - operand_precision - 1`, which is
8 for a product of 68 digits. The case which cancels needs a shift near 34, thus that
function takes its path for two operands which are far apart, and it gives the wrong
result. The 64 bit path works only because its operands have 32 to 38 digits inside a
`u256` of 77.

The product of two 34 digit significands has 67 or 68 digits, and `u256` holds every
value of 77 digits. Both operands move to one exponent, and the width of the pair sets
how far each one goes: the product moves down by 8 digits at the most, and the addend by
38, which is what 128 bits holds. When one of those limits binds, the leading digit of
one operand is more than 8 decades above the leading digit of the other, thus the sum
cannot cancel and the larger operand gives the result. The smaller one then keeps at
least 38 digits below the last digit of the result, where two are sufficient, and a
sticky mark in its last digit holds everything below.

With that correction `decimal128_t` and `decimal_fast128_t` give 100.00% correct results
on the same 20000 triples, for `fma` and for `unchecked_fma`, and 100.00% under full
cancellation.

The correction is not slower than the expression which it replaces. This table measures
the two in one program, on 2000 triples of each shape. The second column is the code of
`develop`, because `d128_fma_impl` is that expression:

| shape | `fma` with the correction, ns | `x * y + z`, ns |
| --- | --- | --- |
| exponents of 10^-3 to 10^3 | 316 | 337 |
| exponents of 10^-40 to 10^40 | 261 | 308 |
| full cancellation | 201 | 137 |
| the addend far above the product | 347 | 278 |
| the addend far below the product | 338 | 289 |
| the addend is zero | 329 | 287 |

On the 20000 triples of the first program the call costs 335 ns with the correction,
against 360 ns for the code of `develop`. Every multiplier and every divisor of the
alignment stays inside 128 bits, thus `u256` takes its narrow paths, and the sum reaches
the constructor as 38 digits with the sticky mark and not as 256 bits.

The split of Veltkamp is in the first program to measure the size of the defect only. It
costs 1698 ns, and its error is as much as one unit of the last place, thus it is not the
correction.

</details>

ibmibmibm:
fma must calculate x * y + z with one rounding, which is the definition of
fusedMultiplyAdd in IEEE 754-2019. The d128_fma_impl function calculated that
expression with the operators of the type, thus it rounded two times. Of 20000
random triples, 16.14% were wrong and the largest error was 118 units of the
last place. When z was -(x * y) the result was 0 for every pair, and the correct
value there is the rounding error of the product. Four functions took that
path: fma and unchecked_fma, of decimal128_t and of decimal_fast128_t.

The 32 and 64 bit paths get their one rounding from a promotion to the next
wider decimal type. The 128 bit types have no wider decimal type, thus this path
holds the exact product of the two significands in 256 bits, it aligns the
addend to that product, and it rounds one time at the end.

The product of two 34 digit significands has 67 or 68 digits, and u256 holds
every value of 77 digits. Both operands move to one exponent, and the width of
the pair sets how far each one goes: the product moves down by 8 digits at the
most, and the addend by 38, which is what 128 bits holds. When one of those
limits binds, the leading digit of one operand is more than 8 decades above the
leading digit of the other, thus the sum cannot cancel and the larger operand
gives the result. The smaller one then keeps at least 38 digits below the last
digit of the result, where two are sufficient, and a sticky mark in its last
digit holds everything below.

Every divisor stays inside 128 bits, thus u256 takes its narrow paths: umul256
for the product and for a move of the addend, div_mod for a move up, and two
steps for a shift of more than 38 digits, because two narrow divisions cost less
than one wide one.

The move up of the product gives its power of ten as a u256. The multiply of a
u256 by a narrow integer reads the words of that integer from its memory. That
order is wrong on a machine which holds the high byte first.

The sum reaches the constructor as 38 digits with the mark, and not as 256 bits.
The constructor does the one rounding, and it gives the same result for every
rounding mode and for a subnormal result, because every digit which the rounding
reads stays above the mark.

decimal128_t and decimal_fast128_t now give 100.00% correct results on the same
20000 triples, for fma and for unchecked_fma, and 100.00% under full
cancellation. This also removes the last 281 inversions of asinh on the 128 bit
types. One call costs 335 ns, against 360 ns for the expression which rounds two
times.

test/test_fma.cpp holds 37 triples with the values of MPFR at 900 bits, the
exact ties of round half to even, which MPFR cannot give because 5e-34 is not a
dyadic rational, the non-finite and the zero cases, the rounding modes, and two
identities over the whole range of the exponent: fma(x, 1, z) is x + z, and
fma(x, y, 0) is x * y.

Fixes #1444

### DIFF
--- a/include/boost/decimal/detail/cmath/fma.hpp
+++ b/include/boost/decimal/detail/cmath/fma.hpp
@@ -11,6 +11,10 @@
 #include <boost/decimal/decimal128_t.hpp>
 #include <boost/decimal/decimal_fast128_t.hpp>
 #include <boost/decimal/detail/config.hpp>
+#include <boost/decimal/detail/i256.hpp>
+#include <boost/decimal/detail/normalize.hpp>
+#include <boost/decimal/detail/power_tables.hpp>
+#include <boost/decimal/detail/u256.hpp>
 
 namespace boost {
 namespace decimal {
@@ -96,10 +100,202 @@ constexpr auto d64_fma_impl(T x, T y, T z) noexcept -> T
     return detail::d128_add_impl_new<T>(first_res, z_components);
 }
 
-template <bool, BOOST_DECIMAL_DECIMAL_FLOATING_TYPE T>
+// 10^38 is the largest power of ten which 128 bits holds. Every divisor of the alignment
+// stays at this width, thus u256 takes its narrow paths.
+BOOST_DECIMAL_INLINE_CONSTEXPR_VARIABLE std::int32_t fma_narrow_digits {38};
+
+// Marks a quotient which dropped a digit that is not zero: a last digit of zero becomes a
+// one. The mark stays far below the last digit of the result, thus the rounding is exact.
+BOOST_DECIMAL_CUDA_CONSTEXPR auto fma_apply_sticky(const u256& quotient, const bool inexact) noexcept -> u256
+{
+    if (inexact && quotient % UINT64_C(10) == u256{0, 0, 0, 0})
+    {
+        return quotient + u256{0, 0, 0, 1};
+    }
+
+    return quotient;
+}
+
+// Divides by a power of ten and reports a dropped digit which is not zero. A shift of more
+// than fma_narrow_digits goes in steps, because a narrow divisor costs less than a wide one.
+BOOST_DECIMAL_CUDA_CONSTEXPR auto fma_shift_right(const u256& sig, std::int32_t shift, const std::int32_t width, bool& inexact) noexcept -> u256
+{
+    if (shift > width)
+    {
+        // Every digit falls below the sticky position, and the value is not zero.
+        inexact = false;
+        return u256{0, 0, 0, 1};
+    }
+
+    u256 quotient {sig};
+    inexact = false;
+
+    while (shift > 0)
+    {
+        const auto step {shift < fma_narrow_digits ? shift : fma_narrow_digits};
+        const auto result {detail::impl::div_mod(quotient, detail::pow10_u128(static_cast<std::size_t>(step)))};
+
+        inexact = inexact || result.remainder != u256{0, 0, 0, 0};
+        quotient = result.quotient;
+        shift -= step;
+    }
+
+    return quotient;
+}
+
+// Narrows the sum to the 38 digits of a 128 bit significand, with the sticky mark in the
+// last place. The constructor then rounds those digits one time.
+BOOST_DECIMAL_CUDA_CONSTEXPR auto fma_narrow_sum(const u256& sig, std::int32_t& exp) noexcept -> int128::uint128_t
+{
+    const std::int32_t digits {detail::num_digits(sig)};
+
+    if (digits <= fma_narrow_digits)
+    {
+        return int128::uint128_t{sig[1], sig[0]};
+    }
+
+    const auto shift {digits - fma_narrow_digits};
+    bool inexact {};
+    const auto shifted {fma_shift_right(sig, shift, digits, inexact)};
+    exp += shift;
+
+    // The quotient holds 38 digits, thus the mark goes on in 128 bits and not in 256.
+    int128::uint128_t narrow {shifted[1], shifted[0]};
+
+    if (inexact && narrow % UINT64_C(10) == 0U)
+    {
+        ++narrow;
+    }
+
+    return narrow;
+}
+
+// fma must calculate x * y + z with one rounding. This path holds the exact product of the
+// two significands in 256 bits, it aligns the addend to it, and it rounds one time.
+template <bool checked, BOOST_DECIMAL_DECIMAL_FLOATING_TYPE T>
 constexpr auto d128_fma_impl(T x, T y, T z) noexcept -> T
 {
-    return x * y + z;
+    #ifndef BOOST_DECIMAL_FAST_MATH
+    BOOST_DECIMAL_IF_CONSTEXPR (checked)
+    {
+        if (!isfinite(x) || !isfinite(y) || !isfinite(z))
+        {
+            return x * y + z;
+        }
+    }
+    #endif
+
+    auto x_components {x.to_components()};
+    auto y_components {y.to_components()};
+
+    // A zero in the product leaves one addition, and that already rounds one time.
+    if (x_components.sig == 0U || y_components.sig == 0U)
+    {
+        return x * y + z;
+    }
+
+    auto z_components {z.to_components()};
+
+    detail::expand_significand<T>(x_components.sig, x_components.exp);
+    detail::expand_significand<T>(y_components.sig, y_components.exp);
+
+    u256 product_sig {detail::umul256(x_components.sig, y_components.sig)};
+    auto product_exp {static_cast<std::int32_t>(x_components.exp + y_components.exp)};
+    const bool product_sign {x_components.sign != y_components.sign};
+
+    if (z_components.sig == 0U)
+    {
+        const auto product_narrow {detail::fma_narrow_sum(product_sig, product_exp)};
+        return T{product_narrow, product_exp, product_sign};
+    }
+
+    detail::expand_significand<T>(z_components.sig, z_components.exp);
+
+    const auto addend_exp {static_cast<std::int32_t>(z_components.exp)};
+    const bool addend_sign {z_components.sign};
+
+    constexpr std::int32_t max_width {76};
+    constexpr std::int32_t sig_dig {detail::precision_v<T>};
+    constexpr std::int32_t product_width {2 * sig_dig};
+
+    // The exponent of the sum. The product must not pass max_width digits on the way, and
+    // the addend must not move by more than 38 digits.
+    auto target_exp {product_exp < addend_exp ? product_exp : addend_exp};
+    const std::int32_t product_floor {product_exp - (max_width - product_width)};
+    const std::int32_t addend_floor {addend_exp - detail::fma_narrow_digits};
+
+    if (target_exp < product_floor)
+    {
+        target_exp = product_floor;
+    }
+    if (target_exp < addend_floor)
+    {
+        target_exp = addend_floor;
+    }
+
+    // The multiplier goes as a u256. A u256 times a narrow integer reads the words of that
+    // integer from its memory, and that order is wrong on a big-endian machine.
+    if (product_exp > target_exp)
+    {
+        product_sig = product_sig * detail::pow10(u256{static_cast<std::uint64_t>(product_exp - target_exp)});
+    }
+    else if (product_exp < target_exp)
+    {
+        bool inexact {};
+        const auto shifted {detail::fma_shift_right(product_sig, target_exp - product_exp, product_width, inexact)};
+        product_sig = detail::fma_apply_sticky(shifted, inexact);
+    }
+
+    // The addend is still 128 bits here. A move down is one multiply into 256 bits, and a
+    // move up is one division of 128 bits.
+    u256 addend_sig {};
+
+    if (addend_exp > target_exp)
+    {
+        addend_sig = detail::umul256(z_components.sig,
+                                     detail::pow10_u128(static_cast<std::size_t>(addend_exp - target_exp)));
+    }
+    else if (addend_exp == target_exp)
+    {
+        addend_sig = u256{z_components.sig};
+    }
+    else if (target_exp - addend_exp >= sig_dig)
+    {
+        // Every digit of the addend falls below the sticky position.
+        addend_sig = u256{0, 0, 0, 1};
+    }
+    else
+    {
+        const auto divisor {detail::pow10_u128(static_cast<std::size_t>(target_exp - addend_exp))};
+        const auto quotient {z_components.sig / divisor};
+        addend_sig = detail::fma_apply_sticky(u256{quotient}, quotient * divisor != z_components.sig);
+    }
+
+    u256 result_sig {};
+    bool result_sign {};
+
+    if (product_sign == addend_sign)
+    {
+        result_sig = product_sig + addend_sig;
+        result_sign = product_sign;
+    }
+    else
+    {
+        // i256_sub gives the magnitude, and it reports which operand was the larger one.
+        const auto product_is_smaller {detail::i256_sub(product_sig, addend_sig, result_sig)};
+        result_sign = product_is_smaller ? addend_sign : product_sign;
+    }
+
+    if (result_sig == u256{0, 0, 0, 0})
+    {
+        // IEEE 754-2019 6.3 asks for a negative zero in the downward mode. The operators and
+        // the 32 and 64 bit paths give a positive zero in every mode, thus this path does too.
+        result_sign = false;
+    }
+
+    const auto result_sig_narrow {detail::fma_narrow_sum(result_sig, target_exp)};
+
+    return T{result_sig_narrow, target_exp, result_sign};
 }
 
 #ifdef _MSC_VER
--- a/test/test_fma.cpp
+++ b/test/test_fma.cpp
@@ -1,0 +1,467 @@
+// Copyright 2026 Matt Borland
+// Distributed under the Boost Software License, Version 1.0.
+// https://www.boost.org/LICENSE_1_0.txt
+//
+// fma must calculate x * y + z with one rounding (IEEE 754-2019 fusedMultiplyAdd).
+// The expected values come from MPFR at 900 bits, at the 34 digits of decimal128_t.
+
+#include <boost/decimal.hpp>
+#include <boost/core/lightweight_test.hpp>
+#include <cstring>
+#include <iostream>
+#include <limits>
+#include <random>
+
+using namespace boost::decimal;
+
+struct fma_case
+{
+    const char* x;
+    const char* y;
+    const char* z;
+    const char* expected;
+};
+
+static const fma_case fma_cases[] =
+{
+// the three worst triples of the report
+    { "4.891128916954585116359252097589272e+03",
+      "-7.570510793333322422709743010426965e-02",
+      "3.707536042839128637930999007415480e+02",
+      "4.701617100187612469012746705158118e-01" },
+    { "9.773367626875150075486696879175602e-02",
+      "-3.337907954067831540381441089089419e-01",
+      "3.235029480287355880201961692407601e-02",
+      "-2.723067369020513807361798191291154e-04" },
+    { "-1.635908948849659232280107483909349e+02",
+      "-3.176446739966658464358952445855542e-03",
+      "-5.281416298488918666313435068678353e-01",
+      "-8.503865103313556434996443805914455e-03" },
+// random triples, exponents of 10^-3 to 10^3
+    { "-1.065008300110031905957897280440974e-03",
+      "8.679786286031742647482469037679398e-01",
+      "2.459772220945470025638742347671538e-01",
+      "2.450528176507664992227350252829746e-01" },
+    { "2.204698774688623268682332486262565e+02",
+      "4.391348326395192744337294837450587e-01",
+      "6.616668522588143528118897915024711e+03",
+      "6.713484525332487707332688687817646e+03" },
+    { "-9.976911973353208647336163725677450e-01",
+      "2.169266061245795773782627422211721e-03",
+      "-7.439068330870974128742758471652387e+00",
+      "-7.441232588524957322211015428677276e+00" },
+    { "-1.888740352881831634403692216579063e+01",
+      "-7.900118649737038306011084617881166e+02",
+      "-4.999754949433110540456096531467187e+01",
+      "1.487127533681834187300900128042427e+04" },
+    { "8.171530766625406270093349670323739e+03",
+      "2.593580490537856789189978389609014e-03",
+      "-5.576232872708608952707412919282934e-01",
+      "2.063589948687864924607130452970817e+01" },
+    { "-7.750345597256329137563280171352258e-02",
+      "-9.100287754281406319757676608288459e+02",
+      "3.073392767787947858334881593235580e+02",
+      "3.778696519089553701144226749142327e+02" },
+    { "6.013317852722187956978891138409258e-03",
+      "1.570765883557867700949175676275650e-01",
+      "-3.229835276084885365519028711089201e-02",
+      "-3.135380130784430692263921963834903e-02" },
+    { "9.033523195572541015573502066498907e+02",
+      "-2.046124559993565407347510599148056e+01",
+      "4.920017356632420627073578236505352e-03",
+      "-1.848370875371517575917133434459618e+04" },
+    { "7.517577000791313867554357371910665e+00",
+      "8.135491485969028894823460725489820e-03",
+      "7.227824998636771650829718582341538e+00",
+      "7.288984182321825972393996547928244e+00" },
+    { "-1.864205891267590352981025441417601e-03",
+      "-4.189531285180449361628988649292932e-03",
+      "-7.773177529299234971239730042481897e+02",
+      "-7.773177451197745936407001508457541e+02" },
+    { "8.739132880157219494689235001232555e+03",
+      "-8.892335805012632280923256547689715e+02",
+      "-4.914910797873741926771937139941189e+00",
+      "-7.771135336409319086085617356230487e+06" },
+    { "-1.396679541697529713725416669038118e+00",
+      "7.849913763434619235676244454514602e+01",
+      "-6.249747111291060839473824917109365e-02",
+      "-1.097006370459028573861129989848511e+02" },
+// random triples, exponents of 10^-40 to 10^40
+    { "5.356601629996288981095041947483554e-06",
+      "-1.107743045106680743270351100745053e+10",
+      "-8.592767617963135435622045638865114e-25",
+      "-5.933738201035498737909347544062264e+04" },
+    { "9.512477724053467805288945525844576e+40",
+      "1.944834564552786098638350181423332e+16",
+      "-4.211117084368772467577488160835992e-02",
+      "1.850019547227760382125089235816492e+57" },
+    { "-2.305296013399484023822372416525407e-19",
+      "7.655711025294320470305879153973029e+13",
+      "-4.447673864198572267900470012730579e+15",
+      "-4.447673864198572267918118692836928e+15" },
+    { "-1.264876497583257991687893341033101e+24",
+      "-3.742990996474187486609778516229004e+34",
+      "-4.353475031420709188973353405088790e-28",
+      "4.734421342105939030493612314372404e+58" },
+    { "-6.229034438331746090377954099392618e+21",
+      "5.878909887578624764747669923672911e-38",
+      "8.356417713218954469061109119733163e-23",
+      "-3.661992379315855424422692169895862e-16" },
+    { "-5.367736664884465261136500864916400e-01",
+      "1.446078880223927545847047179606676e-40",
+      "4.552505687031048811640585814577223e-06",
+      "4.552505687031048811640585814577223e-06" },
+    { "7.825676209954665354533009758092445e+13",
+      "7.348099629599945735417377676212569e-32",
+      "3.227421877809614677045532446929357e+31",
+      "3.227421877809614677045532446929357e+31" },
+    { "-8.765245137926423343038120163995087e+36",
+      "4.738139478573062887311022770079424e-04",
+      "5.258003222290126591065982346675321e-05",
+      "-4.153095402737977818771867378570165e+33" },
+// full cancellation: z = -(x * y)
+    { "-9.821591768821046489724273154673145e+03",
+      "-5.695644916035523607121140304299307e-02",
+      "-5.594029922546193912029752695961763e+02",
+      "-4.960198068695729938168958164989485e-32" },
+    { "-7.837759609545990790735737501578442e-01",
+      "7.360064652792556650772937249901802e+02",
+      "5.768641745930463709291680549854855e+02",
+      "-1.871140972921262543667584500152484e-32" },
+    { "2.577536822883360960346560946174337e+03",
+      "-9.859520508228492288082021687542496e-01",
+      "2.541327716593260836677977625895304e+03",
+      "3.352094880709892407560976878748480e-32" },
+    { "3.830620546279893542174381366992733e+03",
+      "-4.816325532469494111462555739418184e+00",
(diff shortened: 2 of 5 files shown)


include/boost/decimal/detail/cmath/fma.hpp
```diff
@@ -96,10 +100,233 @@ constexpr auto d64_fma_impl(T x, T y, T z) noexcept -> T
...
+// The product of two 34 digit significands has 67 or 68 digits, and u256 holds every
+// value of 77 digits. Both operands move to one exponent, and the width of the pair
+// sets how far each one goes: the product moves down by 8 digits at the most, and the
+// addend by 38, which is what 128 bits holds.
+//
+// When one of those limits binds, the leading digit of one operand is more than 8
+// decades above the leading digit of the other, thus the sum cannot cancel and the
+// larger operand gives the result. The smaller one then moves up, and it keeps at
+// least 38 digits below the last digit of the result, where two are sufficient. A
+// sticky mark in its last digit holds everything below.
```
mborland: Please cut all of the comments down to one or two useful lines. I'm not sure why Claude struggles so much with brevity.

include/boost/decimal/detail/cmath/fma.hpp
```diff
@@ -96,10 +100,233 @@ constexpr auto d64_fma_impl(T x, T y, T z) noexcept -> T
...
+        }
+    }
+    #endif
+
+    auto x_components {x.to_components()};
+    auto y_components {y.to_components()};
+    auto z_components {z.to_components()};
+
+    // A zero in the product leaves one addition, and that already rounds one time.
+    if (x_components.sig == 0U || y_components.sig == 0U)
```
mborland: Currently you could defer decoding z until after this check. It's probably better to defer all of the encodings until after the zero check.

test/test_fma.cpp
```diff
@@ -0,0 +1,460 @@
...
+    BOOST_TEST_EQ(fma(value, one, zero), value);
+    BOOST_TEST_EQ(fma(zero, zero, zero), zero);
+    BOOST_TEST_EQ(fma(value, -one, value), zero);
+}
+
+// The result of a directional mode must hold the correct value between it and the
+// result of the other direction, and the two must be one unit of the last place apart
+// when the exact value does not fit.
+template <typename T>
+void test_rounding_modes()
```
mborland, line 303: This needs to test every rounding mode, not just downward and upward.

mborland (changes requested):

include/boost/decimal/detail/cmath/fma.hpp
ibmibmibm: Done.

test/test_fma.cpp
ibmibmibm: Done.

include/boost/decimal/detail/cmath/fma.hpp
ibmibmibm: Done. All comments rewrote.

mborland (approved):

PR #1446, opened 2026-09-06
Refactor inverse hyperbolic trig funcs

ckormanyos:
Based on improvements from #1432 and #1443, this PR improves the implementations and testing of the inverse hyprebolic trigonometric functions.

### DIFF
--- a/include/boost/decimal/detail/cmath/acosh.hpp
+++ b/include/boost/decimal/detail/cmath/acosh.hpp
@@ -67,42 +67,52 @@ constexpr auto acosh_impl(const T x) noexcept
             // Use (parts of) the implementation of acosh from Boost.Math.
 
             constexpr T root_epsilon { sqrt(std::numeric_limits<T>::epsilon()) };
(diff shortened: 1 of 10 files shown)


include/boost/decimal/detail/cmath/acosh.hpp
```diff
@@ -75,8 +75,32 @@ constexpr auto acosh_impl(const T x) noexcept
...
+                    const auto inv_xsq = one / (x * x);
+
+                    result =
+                        ::boost::decimal::log(x) + numbers::ln2_v<T>
+                      - inv_xsq *
+                        (
+                            one / T { 4, 0 }
+                          + inv_xsq *
+                            (
+                                T { 3, 0 } / T { 32, 0 }
```
mborland: It might be beneficial to make these `constexpr` constants so that we can guarantee that we are not doing this division at runtime.

ckormanyos: Agreed, will do

PR #1464, opened 2026-09-21
Precompile boost/decimal.hpp for the tests

### LINKED ISSUE #1461: CI is timing out
We're starting to break 4 hours of testing. Need to pare things back significantly to get this back under control.

ibmibmibm:
- b2 builds test/pch.hpp once per variant and force-includes it into every test.
- A private feature adds the pch as a source to every target but the pch itself.
- A test which defines a library macro before the include opts out with <pch>off.
- The compile tests of the separate headers opt out as well.
- The opt-out is on a second line, so the CMake test list stays the same.
- GCC pch is off on MinGW, as in Boost.Math.
- test_asinh opts out on clang 6 to 8, which fail its constexpr at C++14 with the pch.
- github_issue_1329 and test_cbrt_from_math opt out: 32-bit GCC 12 loses the -Wstringop-overflow
  pragma of libstdc++ with the pch and then rejects the cpp_dec_float output of Boost.Multiprecision.
- MSVC on an ARM64 host builds without the pch, because it fails to map the pch there now and then.
- Clang 22 needs 4562 s of CPU for the test directory without the pch and 1136 s with it.
- GCC 16 needs 2227 s without and 1442 s with it. Both numbers include the test runs.
- In CI the clang 6 job went from over 180 min to 33 min, MSVC x64 from 118 min to 37 min,
  and clang-win x64 from 248 min to 51 min.

Fixes #1461

---

The first commit is the float compare change:

- The float constructors of the six types tested NaN with != and ±inf with == against
  numeric_limits under a -Wfloat-equal pragma. They now use !(val >= val) for NaN and
  !(val - val >= 0) for ±inf, which give the same answers, and the pragma region is gone.
- compute_float80_128 tests +inf against max() the same way and loses its region too.
- The new test does not read numeric_limits, which is empty for __float128 on libstdc++
  before GCC 16. There the old test made a __float128 zero an infinity and let a
  __float128 infinity through as a finite value. The fast basis tests now cover the zero,
  the infinities and NaN.
- GCC 8 to 14 do not apply the diagnostic pragmas of a precompiled header to the
  templates which they instantiate in the including file, so the pch commit needs this.

The second commit is the rest of the header fix:

- GCC 8 to 14 do not apply the diagnostic pragmas of a precompiled header to the
  templates which they instantiate in the including file. With -Werror every test then
  fails when boost/decimal.hpp is precompiled, which the pch commit does.
- coefficient_rounding converts the shift to unsigned before T1 takes it.
- memcpy into a uint128, an int128 or a bit_cast result goes through a void* (the
  documented way to keep -Wclass-memaccess quiet).
- pow10 tells the optimizer that the index is in range, or GCC 8 to 12 report
  -Warray-bounds on dead paths.
- The whole test directory compiles without a warning on gcc 8 to 16, with and without
  the pch, and on clang 22.

The third commit is the from_bid change:

- Clang 15 to 18 drop an explicit specialization of a constrained function template when
  it comes from a precompiled header, and the from_bid<fast> calls then fail to link.
- The fast type now has its own overload, constrained with is_fast_type_v, and the
  primary is constrained with is_ieee_type_v, so exactly one overload matches.
- The four from_bid tests compile and run at C++14, 17 and 20 with gcc and clang, and
  link with the pch on clang 15 and 16.

The fourth commit fixes a race in the to_from_file example:

- b2 runs the variants of the example in one directory, some of them at the same time,
  and they all used example_values.txt. One variant then read a file which another one
  had removed, and exited with 1.
- The name now comes from argv[0], so every variant has a file of its own. Two copies of
  the old example fail 4 times in 100 parallel runs, the new one never.
- The write side opens the file in binary mode like the read side, which matters on Windows.


mborland:
Hi @ibmibmibm. This looks good, but is currently marked as draft. Was there anything else you were intending to add before making it for review?

ibmibmibm:
It seems that clang 6 to 8 had some issues on constexpr + pch, but i'll mark this as ready anyway.

mborland:
> It seems that clang 6 to 8 had some issues on constexpr + pch, but i'll mark this as ready anyway.

Those are still running in the CI system in under an hour so not a big deal for the purposes of closing out the issue at least.

ibmibmibm:
In `README.md:41-48`, the project still support for GCC 8+ and Clang 6+, so we need to add other method that checks these compiler

mborland:
> In `README.md:41-48`, the project still support for GCC 8+ and Clang 6+, so we need to add other method that checks these compiler

Those are still handled in Drone, just not on GitHub Actions. Boostorg has far more control over Drone so we can make sure those older images don't break

mborland (approved):

